PR #62, opened 2024-06-28
Fix linting and change to npm

fabian-emilius:
- eslint did not check any files before
- eslint typescript is not compatible with eslint 9, so downgraded to 8
- change yarn to npm as requested by Stephan Krusche


krusche (approved):

krusche:
The changes look good 👍 Thanks for improving the config and fixing the linter configuration

PR #131, opened 2024-07-21
Implement basic app shell for the authenticated area

fabian-emilius:
This is a large refactoring PR but needed to set the foundations for the planned features. The old functionality to submit an application and to review application still exists and is working

I am sorry that this PR is so large. This should not be required going forward. Dont be intimidated by the amount of files. Most files are just containing an empty component because I already implemented a component for every page that will exist.

Testing URLs:
- Submit application: https://thesis-track-dev.ase.cit.tum.de/applications/thesis
- Review applications: https://thesis-track-dev.ase.cit.tum.de/management/thesis-applications

Screenshot of new AppShell + Old review application logic
image

Changes
- Implement a basic app shell for the authenticated area where students will be able to submit applications and advisors can manage applications and ongoing theses (src/app/layout/AuthenticatedArea)
- Updated authentication flow using context api because multiple pages need authentication now (src/contexts/AuthenticationContext)
- New request logic: Remove axios and use standard fetch api. This makes it easier to do requests in react with proper cancelling (src/requests/request.ts)
- Remove the need of 2 keycloak client and only use one for client and server. The client can be configured with environment variables now to allow a separated staging setup
- Rename environment variables for easier configuration (webpack.config.js)
- Add empty components for all pages and introduce new folder structure that makes it easier to work on a larger application (src/pages)
- Move old logic into legacy folder and rename all components to legacy that will be replaced later on (src/legacy)
- Add runtime environment variables to make docker images reusable across universities
- Reduce bundle size from 5MB to 800kb (initial load). File splitting is improved and lazy loading for all pages is added (webpack.config.js)
- Remove unnecessary packages to reduce bundle size (package.json)
- Improve build time to sub 10 seconds (webpack.config.js)

Deployment Changes
- Renamed environment variable `REACT_APP_SERVER_HOST` to `API_SERVER_HOST`
- Renamed environment variable `REACT_APP_KEYCLOAK_HOST` to `KEYCLOAK_HOST`
- Renamed environment variable `REACT_APP_KEYCLOAK_REALM_NAME` to `KEYCLOAK_REALM_NAME`
- Added `KEYCLOAK_CLIENT_ID` to configure keycloak client id as environment
- Only one keycloak client should be used going forward. I would suggest creating a new client called `thesis-track-app`. This client should be public and not authenticated. To access the application management, the new `admin` resource role is needed (Create role `admin` in newly created client or use existing thesis-track-client keycloak client). I added an example keycloak realm export to the repository but probably not needed

This new configuration should allow to configure a seperate client for the dev / staging deployment which makes it easier to test features


robertjndw:
Is it really necessary to make the client unauthenticated? As far as I understand: without authentication, any user can access the Keycloak client, potentially leading to unauthorized access to protected resources and sensitive information 🤔

robertjndw (approved):

PR #154, opened 2024-08-09
Allow students and advisors to post comments in thesis section

fabian-emilius:
This PR implements comment function on the thesis page
- Advisors have an own comment section to make comments that are not visible to the student
- Students can see and post comments in the writing section that are visible to everyone that has read access to the thesis

image

### DIFF
--- a/client/src/contexts/ThesisCommentsProvider/ThesisCommentsProvider.tsx
+++ b/client/src/contexts/ThesisCommentsProvider/ThesisCommentsProvider.tsx
@@ -134,6 +134,7 @@ const ThesisCommentsProvider = (props: PropsWithChildren<IThesisCommentsProvider
               ...prev,
               content: prev.content.filter((c) => c.commentId !== comment.commentId),
               totalElements: prev.totalElements - 1,
+              totalPages: Math.ceil((prev.totalElements - 1) / limit),
             }
           })
         } else {
--- a/client/src/contexts/ThesisCommentsProvider/ThesisCommentsProvider.tsx
+++ b/client/src/contexts/ThesisCommentsProvider/ThesisCommentsProvider.tsx
@@ -1,0 +1,147 @@
+import React, { PropsWithChildren, useEffect, useMemo, useState } from 'react'
+import { IThesis, IThesisComment } from '../../requests/responses/thesis'
+import { IThesisCommentsContext, ThesisCommentsContext } from './context'
+import { PaginationResponse } from '../../requests/responses/pagination'
+import { doRequest } from '../../requests/request'
+import { showSimpleError } from '../../utils/notification'
+
+interface IThesisCommentsProviderProps {
+  thesis: IThesis
+  commentType: 'THESIS' | 'ADVISOR'
+  limit?: number
+}
+
+const ThesisCommentsProvider = (props: PropsWithChildren<IThesisCommentsProviderProps>) => {
+  const { children, thesis, commentType, limit = 10 } = props
+
+  const [submitting, setSubmitting] = useState(false)
+  const [comments, setComments] = useState<PaginationResponse<IThesisComment>>()
+
+  useEffect(() => {
+    setComments(undefined)
+
+    return doRequest<PaginationResponse<IThesisComment>>(
+      `/v2/theses/${thesis.thesisId}/comments`,
+      {
+        method: 'GET',
+        requiresAuth: true,
+        params: {
+          commentType,
+        },
+      },
+      (res) => {
+        if (res.ok) {
+          setComments({
+            ...res.data,
+            content: res.data.content.reverse(),
+          })
+        } else {
+          setComments({
+            content: [],
+            totalPages: 0,
+            totalElements: 0,
+            last: true,
+            pageNumber: 0,
+            pageSize: limit,
+          })
+        }
+      },
+    )
+  }, [thesis.thesisId, commentType])
+
+  const contextState = useMemo<IThesisCommentsContext>(() => {
+    return {
+      thesis,
+      comments,
+      posting: submitting,
+      limit,
+      postComment: async (message: string, file: File | undefined) => {
+        setSubmitting(true)
+
+        try {
+          const formData = new FormData()
+          formData.append(
+            'data',
+            new Blob(
+              [
+                JSON.stringify({
+                  message,
+                  commentType,
+                }),
+              ],
+              { type: 'application/json' },
+            ),
+          )
+
+          if (file) {
+            formData.append('file', file)
+          }
+
+          const response = await doRequest<IThesisComment>(
+            `/v2/theses/${thesis.thesisId}/comments`,
+            {
+              method: 'POST',
+              requiresAuth: true,
+              formData,
+            },
+          )
+
+          if (response.ok) {
+            setComments((prev) => {
+              if (!prev) {
+                return prev
+              }
+
+              if (prev.pageNumber > 0) {
+                return prev
+              }
+
+              return {
+                ...prev,
+                content: [...prev.content, response.data].slice(-limit),
+                totalElements: prev.totalElements + 1,
+              }
+            })
+          } else {
+            showSimpleError(`Could not post comment: ${response.status}`)
+          }
+        } finally {
+          setSubmitting(false)
+        }
+      },
+      deleteComment: async (comment: IThesisComment) => {
+        const response = await doRequest(
+          `/v2/theses/${thesis.thesisId}/comments/${comment.commentId}`,
+          {
+            method: 'DELETE',
+            requiresAuth: true,
+          },
+        )
+
+        if (response.ok) {
+          setComments((prev) => {
+            if (!prev) {
+              return prev
+            }
+
+            return {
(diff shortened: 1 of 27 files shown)


client/src/contexts/ThesisCommentsProvider/ThesisCommentsProvider.tsx
```diff
@@ -0,0 +1,154 @@
...
+}
+
+const ThesisCommentsProvider = (props: PropsWithChildren<IThesisCommentsProviderProps>) => {
+  const { children, thesis, commentType, limit = 10 } = props
+
+  const [submitting, setSubmitting] = useState(false)
+  const [page, setPage] = useState(0)
+  const [comments, setComments] = useState<PaginationResponse<IThesisComment>>()
+
+  useEffect(() => {
```
bgeisb, line 21: Consider optimizing the useEffect hook here. Since this effect re-fetches data every time `thesis.thesisId`, `commentType`, `page`, or `limit changes`, it might be worth looking into caching to reduce unnecessary API calls, especially if these values change frequently.

client/src/contexts/ThesisCommentsProvider/ThesisCommentsProvider.tsx
```diff
@@ -0,0 +1,154 @@
...
+              totalElements: prev.totalElements - 1,
+              totalPages: Math.ceil((prev.totalElements - 1) / limit),
+            }
+          })
+        } else {
+          showSimpleError(`Could not delete comment: ${response.status}`)
+        }
+      },
+    }
+  }, [thesis, comments, submitting, commentType, limit, page, setPage])
```
bgeisb, line 145: Is `setPage` needed here?

client/src/pages/ThesisPage/components/ThesisAdvisorCommentsSection/ThesisAdvisorCommentsSection.tsx
```diff
@@ -0,0 +1,38 @@
...
+import ThesisCommentsList from '../ThesisCommentsList/ThesisCommentsList'
+import ThesisCommentsForm from '../ThesisCommentsForm/ThesisCommentsForm'
+
+const ThesisAdvisorCommentsSection = () => {
+  const { thesis, access } = useLoadedThesisContext()
+
+  const [opened, setOpened] = useState(false)
+
+  if (!access.advisor) {
+    return <></>
```
bgeisb, line 14:
```suggestion
    return null
```

client/src/pages/ThesisPage/components/ThesisCommentsForm/ThesisCommentsForm.tsx
```diff
@@ -0,0 +1,52 @@
...
+  return (
+    <Stack>
+      <UploadFileModal
+        title='Attach File'
+        opened={uploadModal}
+        onClose={() => setUploadModal(false)}
+        onUpload={setFile}
+      />
+      <Textarea
+        placeholder='Add a comment'
```
bgeisb, line 24: The placeholder disappears once the user starts typing.
You might consider introducing a label here to improve UX.

client/src/pages/ThesisPage/components/ThesisCommentsList/ThesisCommentsList.tsx
```diff
@@ -0,0 +1,91 @@
...
+          <AuthenticatedFilePreview
+            url={`/v2/theses/${thesis.thesisId}/comments/${openedComment.commentId}/file`}
+            filename={`comment-${openedComment.commentId}`}
+            height={400}
+          />
+        )}
+      </Modal>
+      {!comments &&
+        Array.from(Array(limit).keys()).map((index) => <Skeleton key={index} height={50} />)}
+      {comments?.content.map((comment) => (
```
bgeisb, line 40: When there are no comments to display, consider adding a fallback message indicating that there are no comments yet. This provides better feedback to the user instead of just showing nothing.

bgeisb (changes requested):

client/src/contexts/ThesisCommentsProvider/ThesisCommentsProvider.tsx
fabian-emilius: All of these dependencies are part of the request so yes it is needed in that case as the request should always be up to date with all these states

client/src/pages/ThesisPage/components/ThesisAdvisorCommentsSection/ThesisAdvisorCommentsSection.tsx
fabian-emilius: I will change that

client/src/pages/ThesisPage/components/ThesisCommentsList/ThesisCommentsList.tsx
fabian-emilius: I will add that

bgeisb (approved):

server/src/main/java/thesistrack/ls1/controller/ThesisController.java
```diff
@@ -347,23 +352,97 @@ public ResponseEntity<ThesisDto> deletePresentation(@PathVariable UUID thesisId,
...
+            throw new AccessDeniedException("You need to be a advisor of this thesis to perform this action");
+        }
+
+        if (!comment.getThesis().hasReadAccess(authenticatedUser)) {
+            throw new AccessDeniedException("You do not have the required permissions to view this thesis");
+        }
+
+        return ResponseEntity.ok()
+                .contentType(MediaType.APPLICATION_PDF)
+                .header(HttpHeaders.CONTENT_DISPOSITION, String.format("attachment; filename=comment_%s.pdf", commentId))
```
iyannsch, line 426: Could we use the original filename here? That would allow both advisor and student to know what they are about to download and what they already posted 👍

I'd still recommend a sanitization step that, e.g., only allows `pdf`, `docx`, `tex`, etc

iyannsch (changes requested):
Changes look good 👍 Apart from by change requests, I'd like to bring up the question of whether the users should be able to upload multiple files per comment. I think the filename should persist too

client/src/pages/ThesisPage/components/ThesisCommentsList/ThesisCommentsList.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const [openedComment, setOpenedComment] = useState<IThesisComment>()
+
+  return (
+    <Stack>
+      <Modal opened={!!openedComment} onClose={() => setOpenedComment(undefined)} size='xl'>
+        {openedComment && (
+          <AuthenticatedFilePreview
+            url={`/v2/theses/${thesis.thesisId}/comments/${openedComment.commentId}/file`}
+            filename={`comment-${openedComment.commentId}`}
+            height={400}
```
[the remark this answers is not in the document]
fabian-emilius: The download icon will open a modal that shows the PDF in an iframe with the ability to download the file

server/src/main/java/thesistrack/ls1/controller/ThesisController.java
fabian-emilius: I will create a ticket for that in the backlog. Right now the original filename is not preserved when uploading files to thesis track. Adding that would require larger changes I would not want to do in this PR

client/src/pages/ThesisPage/components/ThesisCommentsList/ThesisCommentsList.tsx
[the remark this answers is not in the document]
fabian-emilius: The idea is more that the comment text tells you what file it is and not the filename

server/src/main/java/thesistrack/ls1/controller/ThesisController.java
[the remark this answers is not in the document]
fabian-emilius: good suggestion!

fabian-emilius:
Thanks for your review @iyannsch . Regarding persisting the filename and allowing to add multiple files to a comment: I will add this to the backlog as I want to keep this feature simple for now and it would require larger changes to allow that at the moment.
I have addressed your other comments

iyannsch (approved):
Looking forward to seeing my other suggestions (especially the "show filename" one) implemented :) Code LGTM now

krusche:
I don't think it's necessary to add multiple files in a comment. I do not really understand why a comment by a student is necessary anyways and why a file upload would be necessary at all. What kind of files would you upload for a thesis as student in the comment section? Let's keep this very simple(!!) and focus on the more important features!!

fabian-emilius:
@krusche The core idea was that the feedback log of the intermediate presentation can be uploaded as comment. Matthias also told me that its sometimes useful that advisors can post comments on the page that are only visible to advisors to keep track of certain important steps / milestones. I thought a simple comment feature would be the easiest solution for that. It probably wont be used that often but can still help to keep track of important messages or files.
Dont worry, the Gantt chart overview will be ready before end of August :)

PR #158, opened 2024-08-10
Implement presentations on thesis page

fabian-emilius:
This PR adds a presentation element to the WRITING section on the thesis page
- Advisors can schedule new presentations
- Advisors can delete scheduled presentations
- Students and viewers can see all scheduled presentations
- Add checkbox whether thesis should be published when adding a final grade

image


server/src/main/java/thesistrack/ls1/service/MailingService.java
```diff
@@ -266,16 +266,21 @@ private Multipart createApplicationFilesMailContent(String text, Application app
...
+        }
 
-        MimeBodyPart cvAttachment = new MimeBodyPart();
-        cvAttachment.attachFile(uploadService.load(application.getUser().getCvFilename()).getFile());
-        multipart.addBodyPart(cvAttachment);
+        String examinationReportFilename = application.getUser().getExaminationFilename();
+        if (examinationReportFilename != null && !examinationReportFilename.isBlank()) {
+            MimeBodyPart examinationReportAttachment = new MimeBodyPart();
+            examinationReportAttachment.attachFile(uploadService.load(examinationReportFilename).getFile());
+            multipart.addBodyPart(examinationReportAttachment);
```
bgeisb, line 280: Similar to line 271/272/273
You might consider extracting this to a shared function.

bgeisb (changes requested):

server/src/main/java/thesistrack/ls1/service/MailingService.java
fabian-emilius: good suggestion

client/src/pages/ThesisPage/components/ThesisWritingSection/components/CreatePresentationModal/CreatePresentationModal.tsx
```diff
@@ -0,0 +1,124 @@
...
+      },
+    },
+  })
+
+  useEffect(() => {
+    form.reset()
+  }, [opened])
+
+  const [creating, onCreatePresentation] = useThesisUpdateAction(async () => {
+    if (!form.values.location && !form.values.streamUrl) {
```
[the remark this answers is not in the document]
fabian-emilius: its a bit tricky to validate 2 inputs that depend on each other with mantine form but I will implement a version of this

bgeisb (approved):
LGTM 👍

server/src/main/java/thesistrack/ls1/service/ThesisService.java
```diff
@@ -265,6 +266,40 @@ public Resource getThesisFile(Thesis thesis) {
...
+        ThesisPresentation presentation = new ThesisPresentation();
+
+        presentation.setThesis(thesis);
+        presentation.setType(type);
+        presentation.setVisibility(visibility);
+        presentation.setLocation(location);
+        presentation.setStreamUrl(streamUrl);
+        presentation.setScheduledAt(date);
+        presentation.setCreatedBy(creator);
+        presentation.setCreatedAt(Instant.now());
```
iyannsch, line 279: Those attributes are not validated here and could be tampered with on client side. In the `ThesisPresentation` class, there are constraints using the `@` notation - do those throw errors when violated? 

You could also get the currently authenticated user for the `createdBy` value instead of accepting the value provided as argument.

server/src/main/java/thesistrack/ls1/service/ThesisService.java
```diff
@@ -324,6 +360,17 @@ public Thesis findById(UUID thesisId) {
                 .orElseThrow(() -> new ResourceNotFoundException(String.format("Thesis with id %s not found.", thesisId)));
     }
 
+    public ThesisPresentation findPresentationById(UUID thesisId, UUID presentationId) {
```
iyannsch, line 363: The naming "find presentation **by ID** suggest that only the `presentationId` is required as a parameter. Why do you need the `thesisId` here?

server/src/main/resources/db/changelog/changes/05_thesis_presentations.sql
```diff
@@ -0,0 +1,4 @@
+--liquibase formatted sql
+
+--changeset emilius:05-thesis-presentations-1
+ALTER TABLE thesis_presentations ADD visibility TEXT NOT NULL DEFAULT 'PRIVATE';
```
iyannsch, line 4: The text is not limited to any length, I think the other migrations had upper bounds for text length, right?

iyannsch (changes requested):
Nice changes! I added a couple of requests 👍

server/src/main/resources/db/changelog/changes/05_thesis_presentations.sql
fabian-emilius: I decided against limiting text length in the database as I dont see any benefit. Under the hood TEXT and VARCHAR(1000) use the same type but just adds a limit check. Modifying accepted string length should not require database modifications and I do that in the backend. I dont have any string limits in the database right now

server/src/main/java/thesistrack/ls1/service/ThesisService.java
fabian-emilius: Its not really required to have it but it prevents security issues because otherwise it could be possible to request presentations of a thesis you dont have access to (if you dont add the check in the calling code).
I prefer to have this check in this method to not forget it when calling it

server/src/main/java/thesistrack/ls1/service/ThesisService.java
fabian-emilius: I have refactored my validation a bit since the last PR:
- I do payload validation in the controller and the service just accepts anything to make it more flexible and usable for different controllers
- The authentication user is also only fetched in the controller. If I would want to get the authenticated user in the service, I would have to pass the JwtToken down to it which I dont want. I think its better if the controller does the authorization and access checks and the service just executes what it should

client/src/pages/ThesisPage/components/ThesisFinalGradeSection/components/SubmitFinalGradeModal/SubmitFinalGradeModal.tsx
```diff
@@ -53,6 +55,11 @@ const SubmitFinalGradeModal = (props: ISubmitFinalGradeModalProps) => {
         />
         <InputLabel required>Feedback</InputLabel>
         <DocumentEditor value={feedback} onChange={setFeedback} editMode={true} />
+        <Checkbox
+          label='Make thesis public'
```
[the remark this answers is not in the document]
fabian-emilius: I have changed it to "Publish Thesis"

fabian-emilius:
@iyannsch thanks for your review. I have addressed your requests or added comments to your suggestion

server/src/main/resources/db/changelog/changes/05_thesis_presentations.sql
iyannsch: Yup sounds good, thanks for the reasoning :) In MySQL, all `TEXT` datatypes are dynamically adjusted in memory utilization so this sounds like a good design 👍

iyannsch (approved):
Code LGTM

server/src/main/resources/db/changelog/changes/05_thesis_presentations.sql
fabian-emilius: Its actually postgres in that case but for postgres TEXT is basically always the better type than character varying

PR #159, opened 2024-08-11
Improve manage application page

fabian-emilius:
Implement "Review Applications v2" with an updated more functional design that requires less clicks to go through applications

image


client/src/pages/ReviewApplicationPage/components/ApplicationListItem/ApplicationListItem.tsx
```diff
@@ -0,0 +1,48 @@
...
+      onClick={onClick}
+      p='md'
+      style={(theme) => ({
+        backgroundColor: selected
+          ? colorScheme === 'dark'
+            ? theme.colors[theme.primaryColor][6]
+            : theme.colors[theme.primaryColor][4]
+          : colorScheme === 'dark'
+            ? theme.colors.dark[6]
+            : theme.colors.gray[1],
```
rabeatwork, line 29: I would outsource that part to make it easier to read.

fabian-emilius:
I will address the different limit on DocumentEditor component in a later PR. Right now the problem is that I want the character limit to be based on text but the form only has limit checks for the html output which is longer than the pure text

PR #195, opened 2024-08-19
Implement reusable Gantt chart

fabian-emilius:
This PR adds a Gantt chart to the thesis overview and dashboard page.

Features
- Theses are grouped by advisor and each group can be collapsed
- The duration of each thesis state is represented in different colors
- The visible time range can be modified with an slider
- Hovering / clicking will show additional information about the thesis

image


client/src/components/ThesesGanttChart/ThesesGanttChart.tsx
```diff
@@ -1,5 +1,153 @@
...
+          [],
+        ),
+      ])
+    : []
+
+  return (
+    <div>
+      <GanttChart
+        columns={[
+          { label: 'Student', width: '120px', textAlign: 'center' },
```
rabeatwork, line 86: Would it make sense to switch to relative values like viewport width instead of sticking to the static pixel values?

rabeatwork:
Two small issues:
1. On the right side, there's a background for a scrollbar, but the scrollbar itself is missing (I hope you understand what I mean). This is a bit distracting.
2. When hovering over the slider, a horizontal scrollbar appears.
Screenshot 2024-08-23 at 13 55 57

client/src/components/GanttChart/GanttChart.module.css
```diff
@@ -0,0 +1,111 @@
...
+}
+
+.headers {
+    display: flex;
+    flex-direction: row;
+    border-bottom: 1px solid var(--mantine-color-text);
+}
+
+.dataHeader {
+    padding: 3px 5px;
```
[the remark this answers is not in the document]
fabian-emilius: yes makes sense

client/src/components/ThesesGanttChart/ThesesGanttChart.tsx
fabian-emilius: Mmh this makes it quite small on mobile. I will change it to a percentage

PR #397, opened 2025-01-09
Rename ThesisTrack to ThesisManagement

robertjndw:
As the system evolved into a more comprehensive thesis management solution, we want to rename it to **ThesisManagement**. This rebranding will make the name more meaningful and better suited for its purpose


fabian-emilius:
PR looks good to me. Maybe it would make sense to rename thesis-track-backup.sh file as well. The PR requires quite some deployment changes though (different network name, new containers etc, but if tested on staging I do not see a problem with that)

Issue #405, opened 2025-01-22
Support multiple groups


krusche:
* There should be a concept of (research) groups that have their own advertised topics. Members of the group automatically create topics within the group. Currently, every advisor / supervisor can see all advertised topics and can also edit them
* When reviewing applications, users can only see the topics in the group unless they are super admin. However, users can currently accept or reject all applications, so this would also require permissions per group
* Currently admins see all theses and applications. There should be something like a group admin, which limits these permissions to the theses the at the group. The current admin should still be available for maintenance reasons, but should become a super admin
* When sending out emails, this should only notify people at the same group (presentations, new application etc)
* Add filter (e.g. selecting a specific group or professor as supervisor) and search options for topics in student applications and the front page
* More complex changes: Each group will have slightly different requirements as to what fields they want in the application. Currently, the fields are rather tailored to a coding heavy group. It should be configurable which data is requested per group in the application
* Each group should have the possibility to write custom email templates

krusche:
Some aspects of this issue are tracked in other issues as well

4ydan:
Any chances getting the feature of adding supervisor/advisors to multiple research groups?

krusche:
This has been implemented. Thesis Management now supports multiple research groups as a first-class concept:

- Users can belong to multiple groups via `AuthenticationContext.researchGroups` (`client/src/providers/AuthenticationContext`)
- Per-group customization is available through `ResearchGroupSettings` (writing guide, presentation slots, automatic rejection, application notification email, etc.)
- Group-scoped administration is exposed via the `group-admin` role and the Research Group Admin / Settings pages

Remaining group-level customization requests (logos, custom mail variables, application field configurability, presentation feature toggle, etc.) are tracked in their own dedicated issues. Closing this umbrella issue as implemented.

PR #436, opened 2025-03-01
fix: `keywords`  can be `NOT NULL`

CommanderStorm:
During evaluating the changes nesssesary to add organisation support (#405), I noticed that the `keywords` field defaults in the database and is null-checked before-insert.
=> this can savely be moved to the default of an empty array

Primary Reason for this PR: Liquibase generates a (slightly misleading) line in the changelog

### DIFF
--- a/server/src/main/java/de/tum/cit/aet/thesis/entity/Thesis.java
+++ b/server/src/main/java/de/tum/cit/aet/thesis/entity/Thesis.java
@@ -61,7 +61,7 @@ public class Thesis {
     private ThesisVisibility visibility;
 
     @JdbcTypeCode(SqlTypes.ARRAY)
-    @Column(name = "keywords", columnDefinition = "text[]")
+    @Column(name = "keywords", columnDefinition = "text[]", nullable = false)
     private Set<String> keywords = new HashSet<>();
 
     @ManyToOne(fetch = FetchType.LAZY)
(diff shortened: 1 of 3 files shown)


krusche:
Thanks for your contribution. After a break, we start the further development again and will review and integrate your changes asap.

server/src/main/java/de/tum/cit/aet/thesis/entity/Thesis.java
```diff
@@ -61,7 +61,7 @@ public class Thesis {
     private ThesisVisibility visibility;
 
     @JdbcTypeCode(SqlTypes.ARRAY)
-    @Column(name = "keywords", columnDefinition = "text[]")
+    @Column(name = "keywords", columnDefinition = "text[]", nullable = false)
```
Claudia-Anthropica: @CommanderStorm [low] Every other NOT NULL field in this entity uses both `@NotNull` (jakarta.validation) and `nullable = false` on `@Column`. For consistency, consider adding `@NotNull` here too — it gives you validation at the JPA layer before hitting the DB.

```suggestion
    @NotNull
    @Column(name = "keywords", columnDefinition = "text[]", nullable = false)
```

In `server/src/main/java/de/tum/cit/aet/thesis/entity/Thesis.java`, the `keywords` field has `nullable = false` on `@Column` but is missing the `@NotNull` annotation that all other NOT NULL fields in this entity use. Add `@NotNull` above the `@Column` annotation on line 64 for consistency.

</details>

CommanderStorm:
I don't want to deal with all of this dehumanising AI-Review crap

PR #482, opened 2025-05-06
Add new error message to display and change logic of displaying uploa…

rampecke:
Trying to upload a presentation as a student, returns now a proper error ('Cannot upload files after final submission. Please contact your advisor.').
The endpoint would already return this error but the client has a extra security layer where it checks the auth before sending out the request resulting in a generic error message. That's why I only added it as edge case to the client.
The logic for getting upload buttons displayed was changed.


marc-fett (approved):
LGTM!

PR #483, opened 2025-05-07
Add small message in final submission and present an Alert Banner for…

rampecke:
A Alert Banner is now shown at the Bottom of the thesis tab, with a link to the official website. The reminder also appears when you mark your submission as final.


marc-fett (changes requested):
Only one small typo

PR #486, opened 2025-05-07
Development: Keycloak Role Bug

marc-fett:
Fixed a bug where new users couldn't see applications, theses and topics because they have no keycloak role yet.


robertjndw (approved):
LGTM 👍🏼

PR #493, opened 2025-05-13
Update client dependencies

krusche:
Update all non-breaking client dependencies


marc-fett (approved):
All components load correctly, no visual or functional regressions noticed after dependency updates.
Build and dev server run fine, Mantine and Tiptap features behave as expected.
No breaking changes observed in application logic.

PR #465, opened 2025-04-13
Development: Filter API calls by research group of user

marc-fett:
### ✨ Summary  

This PR introduces a secure and scalable way to filter all backend API calls by the authenticated user’s assigned research group. It ensures that only authorized users (e.g., advisors, supervisors, professors) can access data from their research group and blocks unauthorized access to other groups’ data. It also replaces the direct injection of @RequestScope beans and improves maintainability across services and controllers.

---

### 🧩What's Included  

- Introduced CurrentUserProvider to manage access control and user context based on the research group
- Filtered access to Thesis, Application, Topic, User, and related entities via centralized access checks
- Updated all controller injections to use ObjectProvider<CurrentUserProvider> instead of directly wiring request-scoped beans
- Migrated all access control logic from the controller level into service-level authorization methods
- Updated client to pass and render research group information where needed
- Updated application, thesis, topic, and presentation services and repositories to enforce research group boundaries
- New postman collection for easy server/endpoint testing

---

#### Controller & Security  
- CurrentUserProvider: central utility to resolve current user, role, and research group scope
- Updated all affected controllers (ApplicationController, UserController, ThesisController, etc.) to use this provider

---

### 📡 New API Endpoints  

| Method | Path                                            | Description                           | Auth Required |
|--------|-------------------------------------------------|---------------------------------------|---------------|
|-|(all relevant endpoints)|now automatically filtered by user's research group|✅|

---

### 🧪 Migration Notes  
- All users (advisor, supervisor, admin) must be assigned to a research group in the database
- Students are not scoped and continue to see all public data
- Admins retain unrestricted access
- If research_group_id is missing on the user, access is denied for scoped roles

---

### 🔮 Next Steps  

- Implement different mail templates for different research groups

PR #495, opened 2025-05-13
Add a overload function to set the state not only to the current Date

rampecke:
This PR allows setting a custom timestamp when saving a ThesisStateChange, instead of always using Instant.now(). It introduces an overloaded method to avoid null checks while keeping the original method unchanged for default behavior.


krusche (approved):
source code changes look good to me and seem to revert a change in #465 which has probably caused the issue

marc-fett (approved):
Tested the changes locally. Everything is working fine. ✅

PR #515, opened 2025-05-21
Added a Admin Panel for ResearchGroups

### LINKED ISSUE #510: Add a ResearchGroupAdmin Page
### Description

With the introduction of the Research Group feature, we need a dedicated **Research Group Admin page** in the UI that allows platform admins to **create, update, and manage research groups and their members**.

This admin interface should include the following capabilities:

---

### Core Requirements

- **Create a new research group**  
  Set group name, description, and assign a head user.

- **Update an existing research group**  
  Change the name, description, or head.

- **Add members to a research group**  
  It must be possible to add users even if they haven't logged into the platform yet.  
  New members should be assigned the role:  
  - `supervisor` if they are the head  
  - `advisor` for all others by default

- **Update member roles**  
  Allow switching between `advisor` and `supervisor` (with `supervisor` being the higher role).

- **Remove members from a group**  
  Provide confirmation before deletion.

- **Searching for members and groups**  
  Add search bars for Members and Research Groups, to make it easier to find them.

rampecke:
Fixes #510


client/src/components/UserInformationRow/UserInformationRow.tsx
```diff
@@ -0,0 +1,44 @@
+import { Group, Text } from '@mantine/core'
+import { ILightUser } from '../../requests/responses/user'
+import CustomAvatar from '../CustomAvatar/CustomAvatar'
+
+type IUserInformationRowProps = {
+  firstName?: string
+  lastName?: string
+  username?: string
+  email?: string
+  disableMessage?: String
```
marc-fett, line 10: please use the primitive here

client/src/components/KeycloakUserAutocomplete.tsx/KeycloakUserAutocomplete.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+interface KeycloakUserAutocompleteProps {
+  selectedLabel: string
+  onSelect: (username: string, label: string) => void
+  label?: string
+  placeholder?: string
+  withAsterisk?: boolean
+}
+
+const KeycloakUserAutocomplete = ({
```
marc-fett, line 27: When the user deletes the user and tries to select the old one again, they can not do it. Please add the possibility to select the old user and/or revert the changes.

client/src/pages/ResearchGroupSettingPage/components/GeneralResearchGroupSettings.tsx
```diff
@@ -0,0 +1,144 @@
...
+      name: researchGroupData?.name || '',
+      abbreviation: researchGroupData?.abbreviation || '',
+      campus: researchGroupData?.campus || '',
+      description: researchGroupData?.description || '',
+      websiteUrl: researchGroupData?.websiteUrl || '',
+      headUsername: researchGroupData?.head.universityId || '',
+    },
+    validateInputOnChange: true,
+    validate: {
+      name: (value) => (value.length < 2 ? 'Name must be at least 2 characters' : null),
```
marc-fett, line 37: Please check for whitespaces too (e.g. use "isNotEmpty")

client/src/pages/ResearchGroupSettingPage/components/GeneralResearchGroupSettings.tsx
```diff
@@ -0,0 +1,144 @@
...
+              maxLength={300}
+              {...form.getInputProps('description')}
+            />
+            <Text size='xs' c='dimmed'>
+              {form.values.description.length}/300 characters
+            </Text>
+          </Grid.Col>
+
+          <Grid.Col span={12}>
+            <Button type='submit' fullWidth mt='md' disabled={!form.isValid()}>
```
marc-fett, line 134: Just two ideas:
1. Disabling the button when no changes were made. 
2. "Discard changes" button

client/src/pages/ResearchGroupSettingPage/components/GeneralResearchGroupSettings.tsx
```diff
@@ -0,0 +1,144 @@
...
+      abbreviation: researchGroupData?.abbreviation || '',
+      campus: researchGroupData?.campus || '',
+      description: researchGroupData?.description || '',
+      websiteUrl: researchGroupData?.websiteUrl || '',
+      headUsername: researchGroupData?.head.universityId || '',
+    },
+    validateInputOnChange: true,
+    validate: {
+      name: (value) => (value.length < 2 ? 'Name must be at least 2 characters' : null),
+      headUsername: (value) => (!value ? 'Please select a group head' : null),
```
marc-fett, line 38: Just an idea: validating website format and showing a warning to help the user

client/src/pages/ResearchGroupSettingPage/components/GeneralResearchGroupSettings.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+          <Grid.Col span={{ base: 12, md: 6 }}>
+            <TextInput label='Abbreviation' {...form.getInputProps('abbreviation')} />
+          </Grid.Col>
+
+          <Grid.Col span={{ base: 12, md: 6 }}>
+            <Select
+              label='Campus'
+              placeholder='Select a campus'
+              data={['Garching', 'Munich', 'Heilbronn', 'Weihenstephan']}
```
marc-fett, line 106: Please add this as constants in the project for better maintainability

client/src/pages/ResearchGroupSettingPage/components/ResearchGroupSettingsCard.tsx
```diff
@@ -0,0 +1,37 @@
+import { Card, Stack, Title, Text } from '@mantine/core'
+import { ReactNode } from 'react'
+
+interface IResearchGroupSettingsCardProps {
+  title: string
+  subtle?: string
+  children: ReactNode
+}
+
+export function ResearchGroupSettingsCard({
```
marc-fett, line 10: The card doesn't look like it is clickable. Please change the cursor and/or add a hover effect

client/src/pages/ResearchGroupSettingPage/components/AddResearchGroupMemberModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { useState } from 'react'
+
+interface IAddResearchGroupMemberModalProps {
+  opened: boolean
+  onClose: () => void
+  researchGroupName: string
+  handleAddMember: (username: string) => void
+}
+
+const AddResearchGroupMemberModal = ({
```
marc-fett, line 12: It is not really clear for the users if they can only select one member or multiple. When I select a user and click again on the same user, I can select again but a strange text is shown (e.g. "admin admin (admin): admin@admin.de -> false" for user "admin admin"). When I click on "Add Member" I get an error in this case.

client/src/pages/ResearchGroupSettingPage/components/AddResearchGroupMember.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+          setMembers((prev) =>
+            prev.map((member) => {
+              if (member.userId === userId) {
+                return { ...member, groups: res.data.groups }
+              }
+              return member
+            }),
+          )
+        } else {
```
marc-fett, line 151: please clear the selected member after getting a response

client/src/pages/ResearchGroupSettingPage/components/AddResearchGroupMember.tsx
```diff
@@ -0,0 +1,266 @@
...
+import { showNotification } from '@mantine/notifications'
+import UserInformationRow from '../../../components/UserInformationRow/UserInformationRow'
+import DeleteButton from '../../../components/DeleteButton/DeleteButton'
+import DeleteMemberModal from './DeleteMemberModal'
+
+interface IAddResearchGroupMemberProps {
+  researchGroupData: IResearchGroup | undefined
+}
+
+const AddResearchGroupMember = ({ researchGroupData }: IAddResearchGroupMemberProps) => {
```
marc-fett, line 34: The name is a bit misleading since this component is for managing the members and not only adding them

client/src/pages/ResearchGroupSettingPage/components/AddResearchGroupMemberModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+  opened,
+  onClose,
+  researchGroupName,
+  handleAddMember,
+}: IAddResearchGroupMemberModalProps) => {
+  const [userDisplayLable, setUserDisplayLabel] = useState('')
```
marc-fett, line 18: Please fix the typo

client/src/pages/ResearchGroupAdminPage/components/CreateResearchGroupModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+export interface ResearchGroupFormValues {
+  name: string
+  abbreviation: string
+  campus: string
+  description: string
+  websiteUrl: string
+  headUsername: string
+}
+
+const CreateResearchGroupModal = ({
```
marc-fett, line 21: Please try to combine the create and edit research group component into one component if possible

client/src/pages/ResearchGroupAdminPage/ResearchGroupAdminPage.tsx
```diff
@@ -0,0 +1,148 @@
...
+          setCreateResearchGroupModalOpened(false)
+          fetchResearchGroups()
+        } else {
+          showSimpleError(getApiResponseErrorMessage(res))
+        }
+      },
+    )
+  }
+
+  return (
```
marc-fett, line 100: please add an information for the user if there are no results for the search query or no research groups exist

client/src/pages/ResearchGroupAdminPage/ResearchGroupAdminPage.tsx
```diff
@@ -0,0 +1,148 @@
...
+            pageSize: -1,
+          })
+        }
+        setResearchGroupsLoading(false)
+      },
+    )
+  }
+
+  useEffect(() => {
+    fetchResearchGroups()
```
marc-fett, line 64: please check if the user really changed the search query to avoid unnecessary api calls. E.g. when the user adds a whitespace, no call is needed.

client/src/components/UserInformationRow/UserInformationRow.tsx
```diff
@@ -0,0 +1,44 @@
...
+  email,
+  disableMessage,
+  user,
+}: IUserInformationRowProps) => {
+  return (
+    <Group gap='xs' wrap='nowrap' align='center' justify='flex-start'>
+      {user && <CustomAvatar user={user} size={30} />}
+
+      <div>
+        <Group gap='xs'>
```
marc-fett, line 26: please add checks for null

client/src/components/KeycloakUserAutocomplete.tsx/KeycloakUserAutocomplete.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+    doRequest<KeycloakUserElement[]>(
+      '/v2/users/keycloak-users',
+      {
+        method: 'GET',
+        requiresAuth: true,
+        params: { searchKey: debouncedSearchKey },
+      },
+      (res) => {
+        setLoadingUsers(false)
```
marc-fett, line 69: please move this to the end

server/src/main/java/de/tum/cit/aet/thesis/service/ResearchGroupService.java
```diff
@@ -118,10 +129,33 @@ public ResearchGroup createResearchGroup(
     return savedResearchGroup;
   }
 
+  private User getUserByUsernameOrCreate(String username) {
```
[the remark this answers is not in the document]
rampecke: Not sure if this makes sense.
I thougth about it, but this is a helper function that is called only by the ResearchGroupService and therefore is private. If I move it then I have to make it public. As of right now I don't want any other class to call that function. 

I get that content-wise it would make sense, but separation of concerns-wise I am not so sure.

You still want me to move it?

client/src/pages/ResearchGroupAdminPage/ResearchGroupAdminPage.tsx
rampecke: Disagree here.
I added a debounch search.
Means it only searches after a specific time when there was no change. I used a value that is not that little. 

Most people don't add only a whitespace, and then since they added a another character the debounce won't search, already minimizing the amount of API calls. Adding a check here would just add unnecessary code in my opinion. It gets called each time and only very rarely make the decision to not make an API call.

Are you sure I should add it?

server/src/main/java/de/tum/cit/aet/thesis/service/ResearchGroupService.java
```diff
@@ -118,10 +129,33 @@ public ResearchGroup createResearchGroup(
...
+      newUser.setJoinedAt(currentTime);
+      newUser.setUpdatedAt(currentTime);
+
+      // Load user data from Keycloak
+      AccessManagementService.KeycloakUserInformation userElement = accessManagementService.getUserByUsername(username);
+
+      newUser.setUniversityId(userElement.username());
+      newUser.setFirstName(userElement.firstName());
+      newUser.setLastName(userElement.lastName());
+      newUser.setEmail(userElement.email());
```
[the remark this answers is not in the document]
rampecke: I'm actually not sure what you want me to do if the email from Keycloak is not valid.

We still have to create the user account even if the email is invalid, since users can only update their email in TUMonline—not in our platform. I could set the email field to NULL, but that would lead to errors messages as well, just more confusing ones.

When the user logs in, the email will be overwritten with the value from Keycloak anyway, as we don’t perform any validation there either.

The alternative would be to also validate the email at login and set it to NULL if it's invalid. However, preventing login because of an invalid email isn't really an option.

So, isn't it better to allow login and ensure the system properly displays the relevant error messages, instead of blocking access or causing unrelated errors?
I mean the error messages you are talking about are there for exactly the reason they should be there, or am I mistaken?

server/src/main/java/de/tum/cit/aet/thesis/service/ResearchGroupService.java
```diff
@@ -160,8 +223,9 @@ public void archiveResearchGroup(ResearchGroup researchGroup) {
     researchGroupRepository.save(researchGroup);
   }
 
-  public void assignUserToResearchGroup(UUID userId, UUID researchGroupId) {
-    User user = userService.findById(userId);
+  public User assignUserToResearchGroup(String username, UUID researchGroupId) {
```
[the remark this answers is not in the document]
rampecke: does exist

server/src/main/java/de/tum/cit/aet/thesis/service/ResearchGroupService.java
```diff
@@ -173,6 +237,70 @@ public void assignUserToResearchGroup(UUID userId, UUID researchGroupId) {
...
+
+        if (!user.getResearchGroup().getId().equals(researchGroupId)) {
+          throw new AccessDeniedException("User is not assigned to this research group.");
+        }
+        if (user.getResearchGroup().isArchived()) {
+            throw new AccessDeniedException("Cannot remove user from an archived research group.");
+        }
+        if (user.getResearchGroup().getHead() == user) {
+            throw new AccessDeniedException("Cannot remove the head of the research group.");
+        }
```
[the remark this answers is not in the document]
rampecke: This was asked to Stephan. (I was also unsure what should happen if I delete someone with open Topics and running Theses)
We don't change it. The new head has to decide individually afterwards. I checked if it is possible for supervisors to change these and topics if they are not by them and they can.

I want to improve this in the future by a follow up ticket, but I would have to change a lot of endpoints again for this and didn't want to bloat up this PR.

server/src/main/java/de/tum/cit/aet/thesis/service/ResearchGroupService.java
```diff
@@ -173,6 +237,70 @@ public void assignUserToResearchGroup(UUID userId, UUID researchGroupId) {
...
+        User user = userService.findById(userId);
+        ResearchGroup researchGroup = findById(researchGroupId);
+
+        if (!user.getResearchGroup().getId().equals(researchGroup.getId())) {
+            throw new AccessDeniedException("User is not assigned to this research group.");
+        }
+
+        if (user.getResearchGroup().isArchived()) {
+            throw new AccessDeniedException("Cannot update role of a user in an archived research group.");
+        }
```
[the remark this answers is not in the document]
rampecke: Since you allowed it in your endpoint 
I decided to also allow it

Groups can decide that the head not be a supervisor if he does not want to be. Maybe there is a group where they have multiple supervisors, but the head does not want to supervise or something. Don't see a harm in allowing changing it here. Won't be used in most cases, but who knows.

server/src/main/java/de/tum/cit/aet/thesis/controller/ResearchGroupController.java
```diff
@@ -67,13 +70,27 @@ public ResponseEntity<ResearchGroupDto> getResearchGroup(
     return ResponseEntity.ok(ResearchGroupDto.fromResearchGroupEntity(researchGroup));
   }
 
+    @GetMapping("/{researchGroupId}/members")
+    @PreAuthorize("hasRole('admin')")
```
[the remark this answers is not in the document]
rampecke: This will change in the future when we allow Research Groups to use the Setting Page for their own Research Group.
But right now, only admins are allowed to show it.

server/src/main/java/de/tum/cit/aet/thesis/service/AccessManagementService.java
```diff
@@ -157,18 +307,36 @@ private UUID getGroupId(String groupName) {
                 .orElseThrow(() -> new RuntimeException("Group not found: " + groupName));
     }
 
-    private record UserElement(UUID id) {}
+    private record ClientElement(UUID id, String clientId, String name) {}
+    private UUID getServiceClientUUID() {
```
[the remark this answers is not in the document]
rampecke: Do you have a better name suggestion?

I also didn't find it a good name, but couldn't come up with something else.

server/src/main/java/de/tum/cit/aet/thesis/service/AccessManagementService.java
```diff
@@ -178,7 +346,31 @@ private UUID getUserId(String username) {
...
+                    )
+                    .headers(headers -> headers.addAll(getAuthenticationHeaders()))
+                    .retrieve()
+                    .bodyToFlux(KeycloakUserInformation.class)
+                    .collectList()
+                    .block();
+            return users;
+        } catch (RuntimeException exception) {
+            log.warn("Could not fetch users from keycloak", exception);
+            return new ArrayList<>();
```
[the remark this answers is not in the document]
rampecke: It is intended that we get an empty list in that case
I can't throw an error and return an empty list

Do you have a better solution?

server/src/main/java/de/tum/cit/aet/thesis/service/AccessManagementService.java
```diff
@@ -106,6 +127,135 @@ private void removeKeycloakGroup(UUID userId, UUID groupId) {
...
+    }
+
+    private void assignKeycloakRole(UUID userId, String role) {
+        if (userId == null || role == null) {
+            throw new RuntimeException("User id or role is null");
+        }
+
+        Role roleObject = getClientRoleByName(role);
+
+        webClient.post()
```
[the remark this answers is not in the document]
rampecke: I am not sure what your talking about.
I was not able to find any .method(Http.POST) in the entire project

I now removed all the .method(Http.somethings) I added. But since the code already has both all over I am not sure if this needs refactoring.

client/src/pages/ResearchGroupSettingPage/components/GeneralResearchGroupSettings.tsx
rampecke: used .trim

client/src/pages/ResearchGroupSettingPage/components/GeneralResearchGroupSettings.tsx
rampecke: I consider it as follow up ticket.

client/src/components/UserInformationRow/UserInformationRow.tsx
rampecke: Solved by default values

server/src/main/java/de/tum/cit/aet/thesis/service/AccessManagementService.java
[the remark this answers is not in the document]
marc-fett: maybe getApplicationClientUUID?

server/src/main/java/de/tum/cit/aet/thesis/service/AccessManagementService.java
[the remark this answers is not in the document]
marc-fett: But why would you like to return anything? Can't the server just throw the error and the user knows, that there are currently problems, so they will try again later.

server/src/main/java/de/tum/cit/aet/thesis/service/AccessManagementService.java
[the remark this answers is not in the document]
marc-fett: This file used mostly the ".method(HttpMethod.XXX)" pattern. I'm fine with both approaches (.get or .method(...)). It would be good to have a unified approach in the whole application at some point in the future. But there are definitely many more important things :D

server/src/main/java/de/tum/cit/aet/thesis/service/AccessManagementService.java
```diff
@@ -26,27 +29,45 @@ public class AccessManagementService {
...
         this.serviceClientSecret = serviceClientSecret;
+        this.clientId = clientId;
 
         this.webClient = WebClient.builder()
                 .baseUrl(keycloakHost)
                 .build();
 
+        UUID clientUUID = null;
+        try {
+            clientUUID = clientId.isBlank() || serviceClientSecret.isBlank() ? null : getServiceClientUUID();
```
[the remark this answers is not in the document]
rampecke: We need this to authenicate against Keycloak Api

PR #799, opened 2026-01-30
Feature/interview management

rampecke:


spethso:
Applicants can still be invited to an interview even if their flag is set to "not interested," which seems odd to me.

spethso:
When selecting the thesis language for the application, the value is not stored for the interview process and accepting students there.

image

MaximilianAnzinger:
When accepting the applicant after the interview, the modal is not closing.
image

And you're thus able to accept multiple times.
image

PR #803, opened 2026-02-09
Sync matriculation number from Keycloak JWT

krusche:
## Summary

This PR adds matriculation number syncing from Keycloak via two complementary mechanisms:

### 1. Primary mechanism: JWT sync on login (permanent)
- Extracts the `matrikelnr` claim from the Keycloak JWT in `AuthenticationService.updateAuthenticatedUser()` on every login, using the same null-guard pattern as email/firstName/lastName
- This is the **long-term solution** — every user gets their matriculation number synced automatically when they log in

### 2. Backfill: Scheduled sync task (temporary)
- `MatriculationNumberSync` fills in matriculation numbers for existing users who were created before the JWT sync was implemented, without requiring them to re-login
- Runs 10 minutes after server start and then every 7 days
- **This task is temporary** and can be disabled once all affected users have been backfilled. The code is kept in place in case a similar backfill is needed in the future.

### 3. KeycloakUserDto: Matriculation number in admin user lookup
- Includes `matriculationNumber` in `KeycloakUserDto` and the frontend `IKeycloakUserElement` interface
- This ensures that users added directly through the Keycloak user search (e.g. when assigning supervisors, advisors, or students to research groups) also get their matriculation number populated immediately, without having to wait for their first login
- Adds `attributes` field to `KeycloakUserInformation` and requests the full user representation (`briefRepresentation=false`) so custom attributes like `matrikelnr` are available

### Other changes
- Fix a potential NPE in `LightUserDto.fromUserEntity()` when `email` is null

## Keycloak Configuration

A protocol mapper must be added to the Keycloak client to expose `matrikelnr` in the JWT:

1. **Clients** > `thesis-management-app` > **Client scopes** tab
2. Click the dedicated scope (e.g. `thesis-management-app-dedicated`)
3. **Mappers** tab > **Configure a new mapper** > **User Attribute**
4. Configure: Name=`Matrikelnummer`, User Attribute=`matrikelnr`, Token Claim Name=`matrikelnr`, Claim JSON Type=`String`, Add to ID/access/userinfo/introspection tokens=ON, Multivalued=OFF

This has already been configured for both https://thesis-dev.aet.cit.tum.de/ and https://thesis.aet.cit.tum.de/.

## Test plan

- [x] Log in as a student with a matriculation number in LDAP and verify the JWT contains the `matrikelnr` claim
- [x] Check `GET /api/v2/user-info` includes the matriculation number
- [x] Verify the Keycloak user search endpoint (`GET /api/v2/users/keycloak?search=...`) returns the matriculation number
- [x] Add users via research group management and verify their matriculation numbers are populated from Keycloak
- [x] Log in as a professor (no matrikelnr) and verify existing data is not cleared
- [x] Run existing and new server tests: `cd server && ./gradlew test`
- [x] Verify the scheduled `MatriculationNumberSync` backfills missing matriculation numbers after server restart (15/19 users updated, 4 remaining are professors/service accounts without matrikelnr in Keycloak)

### DIFF
--- a/client/src/components/UserInformationForm/UserInformationForm.tsx
+++ b/client/src/components/UserInformationForm/UserInformationForm.tsx
@@ -40,6 +40,7 @@ const UserInformationForm = (props: IUserInformationFormProps) => {
 
   const form = useForm<
     Omit<IUpdateUserInformationPayload, 'enrolledAt'> & {
+      matriculationNumber: string // read-only, synced from Keycloak, not sent to server
       semester: string
       researchGroupName: string | null
       customData: Record<string, string>
@@ -283,7 +284,6 @@ const UserInformationForm = (props: IUserInformationFormProps) => {
           <Tooltip label='This field is not editable because it is synchronized from your university account.'>
             <TextInput
               type='text'
-              required={requireCompletion}
               placeholder='Matriculation Number'
               label='Matriculation Number'
               {...form.getInputProps('matriculationNumber')}


client/src/requests/responses/keycloakUser.ts
```diff
@@ -4,5 +4,6 @@ export interface IKeycloakUserElement {
   firstName: string
   lastName: string
   email: string
+  matriculationNumber: string | null
```
rampecke: The IKeycloakUserElement is only relevant for the Keycloak Search in Member Management, it should not be included in this fix, since it is independent from the Problem.

server/src/main/java/de/tum/cit/aet/thesis/cron/MatriculationNumberSync.java
```diff
@@ -0,0 +1,87 @@
...
+ * However, existing users who registered before this sync was implemented may have a {@code null}
+ * matriculation number. This task periodically queries Keycloak's admin API to fill in the gaps
+ * without requiring those users to re-login.</p>
+ *
+ * <p>The task runs 10 minutes after server start and then every 7 days. User IDs are loaded once,
+ * then each user is processed and committed individually so that a single Keycloak failure does not
+ * roll back progress and no long-running transaction is held open during blocking HTTP calls.</p>
+ */
+@Component
+public class MatriculationNumberSync {
```
rampecke: The Cron Job will run forever unless deleted, just to fill holes from a Bug. It would make more sense to make a Script that fetches and updates the data in the database and from then on it should work again. The data should be synced on Login.

The Cron Job is the reason why the PR seems to fix the issue. It uses logic that was not intended to fetch information about a single person and updates it here. It works to update it once per day like that. But it ignores the real Logic that happens in the Authentication Service at Login. But this is problematic since it interferes with the actual usage of the KeycloakUser and the endpoint.

Also, this will not always work since there is a period between the User logging in and the next time the Cron Job runs.

rampecke (commented):
I had a look into the Issue this Ticket is trying to fix. 

The real function that should be doing this is: 

public User updateAuthenticatedUser(JwtAuthenticationToken jwt) in AuthenticationService.

This line 
String matriculationNumber = (String) attributes.get("matrikelnr"); 
Returns null. This should be contained in the JWT Token, but it does not -> Needs to be fixed in Keycloak. No code required.

server/src/main/java/de/tum/cit/aet/thesis/cron/MatriculationNumberSync.java
krusche: I understand the concern! To clarify: the cron job is **not** the primary fix. The actual fix is the JWT-based sync in `AuthenticationService.updateAuthenticatedUser()`, which extracts the `matrikelnr` claim on every login — this is the permanent mechanism. The cron job is only a **temporary backfill** for existing users who were created before the JWT sync existed and may not log in frequently. It can be disabled once the backfill is complete. I've added documentation in the latest commit to make this clearer.

rampecke (approved):

PR #931, opened 2026-03-19
`Development`: Fix server test for ThesisAnonymizationService

bensofficial:
- The `notifiesForThesisApproachingExpiry` test used a 1900-day offset (~5.2 years) for the end date. Until recently, this resulted in an end date in late 2020. Since theses are anonymized at the end of each year after 5 years, the retention expiry (end of that year + 5 years) was Dec 31, 2025, which is why the test previously passed on `develop`.
- As time progressed, the computed end date shifted into early 2021, moving the retention expiry to Dec 31, 2026. This caused the test to fail.
- Increased the offset to 2250 days (~6.3 years) to ensure the retention expiry is always safely in the past, preventing the test from becoming flaky as time advances.

## Summary by CodeRabbit

* **Tests**
  * Adjusted a test for anonymization notification timing by moving historical timestamps further back and updating related comments; test behavior and assertions remain unchanged (still verifies anonymization notification is set).

### DIFF
--- a/server/src/test/java/de/tum/cit/aet/thesis/service/ThesisAnonymizationServiceTest.java
+++ b/server/src/test/java/de/tum/cit/aet/thesis/service/ThesisAnonymizationServiceTest.java
@@ -377,8 +377,8 @@ class ThesisAnonymizationServiceTest extends BaseIntegrationTest {
 		void notifiesForThesisApproachingExpiry() throws Exception {
 			createTestEmailTemplate("THESIS_ANONYMIZATION_REMINDER");
 
-			// Thesis finished ~6.3 years ago → retention expiry at least 5 years ago
-			// which is in the past and thus within the 30-day notification horizon.
+			// Thesis finished ~6.3 years ago → retention expiry is safely in the past
+			// and thus within the 30-day notification horizon.
 			// This thesis should definitely be notified.
 			Instant createdAt = Instant.now().minus(2300, ChronoUnit.DAYS);
 			Instant endDate = Instant.now().minus(2250, ChronoUnit.DAYS);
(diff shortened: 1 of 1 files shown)


server/src/test/java/de/tum/cit/aet/thesis/service/ThesisAnonymizationServiceTest.java
```diff
@@ -377,11 +377,11 @@ class SendAnonymizationNotifications {
 		void notifiesForThesisApproachingExpiry() throws Exception {
 			createTestEmailTemplate("THESIS_ANONYMIZATION_REMINDER");
 
-			// Thesis finished ~5.2 years ago (endDate in 2020) → retention expiry = Dec 31, 2025
+			// Thesis finished ~6.3 years ago → retention expiry at least 5 years ago
```
Claudia-Anthropica: @bensofficial [nit] The comment says "retention expiry at least 5 years ago" but with 2250 days from today the endDate lands in late 2019, so expiry = Dec 31, 2024 — about 1.2 years ago, not 5. Something like "retention expiry is safely in the past" would be more accurate. Totally minor though, not blocking.

In `server/src/test/java/de/tum/cit/aet/thesis/service/ThesisAnonymizationServiceTest.java`, line 380, the comment says "retention expiry at least 5 years ago" which is inaccurate. Change it to "retention expiry is safely in the past" to avoid confusion about the actual time delta.

</details>

Claudia-Anthropica (approved):
@bensofficial Clean fix, the math checks out. The increased offset keeps the retention expiry safely in the past regardless of year boundaries for a long time. Nice catch on the flaky test.

Claudia-Anthropica (approved):
@bensofficial Comment fix looks good, thanks for addressing the nit. Still approving — clean fix overall.

Kiara65 (approved):
Code looks good and the test is not failing anymore, thank you! :)

Issue #827, opened 2026-02-16
US date format is ambiguous (MM/D/YY)


mschmidm:
The application currently uses US date format: `MM/D/YY` → `03/2/26`

<img height="200" alt="Current date format showing 03/2/26" src="https://github.com/user-attachments/assets/57a65179-bc9a-4256-84da-5bcd35ecf517" />

This format is ambiguous for European users. For example, France and the UK use DD/MM/YYYY → 02/03/2026, making it unclear whether 03/2/26 means March 2nd or February 3rd.

**Suggested solutions:**
- **Preferred:** `DD MMM YYYY` → `16 Feb 2026` (universally unambiguous)
- **Alternative:** `YYYY-MM-DD` → `2026-02-16` (ISO 8601 standard)

Thank you for your work! :)

mschmidm:
Thanks for the quick reply! Ah, you are right, my system and browser language is English :D

Nevertheless, I still think that it would be sensible to change to date format for everyone.

The whole application is not localized, therefore it is unexpected that this one single data point is localized. While german users will directly recognize their own date format, users with a french or english browser cannot be sure what format is being used.

Using a format like "2026-02-12" or "12 Feb 2026" for everyone would clarify this.

Another reasonable option, would be to still localize the date format for everyone, but use a date style that is not ambiguous like "medium" or "long":

```javascript
const date = new Intl.DateTimeFormat(undefined, { dateStyle: "medium" }).format(new Date("2026-02-12"));
```

With "medium", this returns "Feb 12, 2026" for English and "12.02.2026" for German.

What do you think about this? :)

PR #950, opened 2026-04-22
feat: normalize all date and time rendering to YYYY-MM-DD / HH:mm

bassner:
## Summary

Replace the inconsistent mix of locale-dependent date formats (US `MM/DD/YYYY`, German `DD.MM.YYYY`, long-form `April 21, 2026`, raw ISO `Instant.toString()`) with a single, locale-neutral, unambiguous layout everywhere a date is shown to a user:

- **Dates:** `YYYY-MM-DD`
- **Times:** `HH:mm` (24h)
- **Date + time (web UI / PDFs):** `YYYY-MM-DD HH:mm`
- **Date + time (outbound emails only):** `YYYY-MM-DD HH:mm (ZONE)` — zone suffix added because recipients may not share the server's timezone

Timezone handling is unchanged (server: `ZoneId.systemDefault()`, client: browser local). Only display format changes. API payloads, iCalendar (`.ics`), and persisted data are untouched.

## What changed

**Client (31 files)**
- `client/src/utils/format.ts`: replaced `formatDate` with three explicit helpers (`formatDate` / `formatTime` / `formatDateTime`), no options flag. Output is built from `getFullYear/getMonth/getDate/getHours/getMinutes` with zero-padding — fully locale-independent. Invalid `Date` inputs → empty string; years <1000 are zero-padded to 4 digits.
- Every `formatDate(...)` call site migrated per intent (date-only vs date+time).
- Direct `toLocaleDateString` / `toLocaleTimeString` / `toLocaleString` and `dayjs().format(...)` renderers migrated to the new helpers.
- All Mantine `DateInput` and `DateTimePicker` instances now set `valueFormat="YYYY-MM-DD"` or `"YYYY-MM-DD HH:mm"` so pickers display (and parse typed input in) the new format.
- Weekday / month-name label extractions and internal state keys (`toDateString()` / `toISOString()` used as map keys) are intentionally left alone.

**Server (10 files, 1 new)**
- `DataFormatter.formatDate` → `yyyy-MM-dd`. `DataFormatter.formatDateTime` → `yyyy-MM-dd HH:mm (zzz)` pinned to `Locale.ENGLISH` so German JVMs don't render `MEZ`/`MESZ`.
- `PDFBuilder.DATETIME_FORMATTER` → `yyyy-MM-dd HH:mm` (no zone suffix — internal document).
- `ThesisAnonymizationService.DATE_FORMATTER`, `UserDeletionService.DATE_FORMATTER` → `yyyy-MM-dd`.
- `ThesisService` assessment PDF: `presentation.scheduledAt` now renders as date+time (previously dropped the time entirely).
- Example strings in `MailInterviewSlot` / `MailThesisPresentation` / `MailApplication` `templateVariables()` updated so the variable-picker UI hints match actual output.
- New Liquibase changeset `40_normalize_date_format_in_rejection_reminder.sql` rewrites the inline Thymeleaf `#temporals.format(app.rejectionDate, 'dd.MM.yyyy')` pattern stored in the `email_templates` table row seeded by migration 14. Migration 14's source is left untouched so Liquibase checksums on existing deployments remain valid.
- `DataFormatterTest` gains a locale-independence test that forces `Locale.GERMAN` + `Europe/Berlin` and asserts the output ends with `(CET)` and does **not** contain `MEZ` — this will fail if someone removes `.withLocale(Locale.ENGLISH)` from production.

## Why

Locale-dependent rendering has led to real misunderstandings (e.g. `04/05/2026` interpreted as April 5 vs May 4). A single structurally-unambiguous format removes the ambiguity without requiring per-user locale preferences.

## Test plan

- [x] `cd client && npx tsc --noEmit` — clean
- [x] `cd client && npx eslint src/` — clean
- [x] `cd server && ./gradlew spotlessApply test` — 752 tests, 751 passed, 1 skipped (pre-existing), 0 failed
- [ ] Playwright E2E suite
- [ ] Manual verification: Theses table, Applications table, Presentations overview, Interview booking, Thesis page — all dates render as `YYYY-MM-DD` / `YYYY-MM-DD HH:mm`
- [ ] Manual verification: typed `YYYY-MM-DD` and `YYYY-MM-DD HH:mm` input works in each migrated `DateInput` / `DateTimePicker`
- [ ] Manual verification: trigger a test interview invitation and presentation invitation via Mailpit; confirm email renders `2026-04-21 14:30 (CEST)` (or local zone equivalent)
- [ ] Manual verification: download a thesis assessment PDF; confirm all dates render `yyyy-MM-dd` and the presentation date renders as `yyyy-MM-dd HH:mm` with no zone suffix

## Notes

Research-group-customized email templates where someone hand-typed `dd.MM.yyyy` into a `#temporals.format(...)` call will not be auto-migrated — those would need manual re-editing. Flag in release notes if relevant.

## Release Notes

* **Refactor**
  * Standardized date and time display formats across the application to ISO format (`YYYY-MM-DD` for dates, `YYYY-MM-DD HH:mm` for timestamps) for improved consistency and clarity.

* **Chores**
  * Updated server-side date formatting utilities and email template examples to reflect new standard date format.
  * Updated database migrations to normalize date formats in email templates.


Claudia-Anthropica (approved):
@bassner Clean, thorough normalization across the full stack. The timezone-safe parsing for date-only strings and the locale-pinned zone abbreviations are nice touches. Good call adding the Liquibase migration for the stored template rather than editing the original changeset.

client/src/utils/format.ts
```diff
@@ -7,32 +7,64 @@ import { GLOBAL_CONFIG } from '../config/global'
...
   }
 
   const item = new Date(date)
+  if (Number.isNaN(item.getTime())) {
+    return null
+  }
+  return item
+}
+
+export function formatDate(date: string | Date | null | undefined): string {
```
Claudia-Anthropica, line 46: @bassner [high] The develop merge dragged in `client/test/format.test.mjs` (the #966 guard test), and its `formatDate — issue #827` describe block pins the *old* implementation — it reads this very source file and asserts `dateStyle: 'medium'` is present plus the `withTime ? { timeStyle: 'short' } : {}` gate. Since you replaced all of that with the manual ISO builder, the `node --test test/*.test.mjs` step fails and reds `run-tests / build`. That whole describe block needs rewriting for the new format (the `ensureAbsoluteLinkHref` block below it is fine and should stay).

In `client/test/format.test.mjs`, the `describe('formatDate — issue #827 ...')` block (roughly lines 14-99) pins the obsolete `dateStyle: 'medium'` implementation and fails because `client/src/utils/format.ts` no longer uses `toLocaleString`. Rewrite that block to match the new helpers: assert `formatDate(...)` yields `YYYY-MM-DD`, `formatTime(...)` yields `HH:mm`, and `formatDateTime(...)` yields `YYYY-MM-DD HH:mm`, and replace the source-pinning assertions (`/dateStyle:\s*'medium'/`, the `withTime` ternary regex) with checks for the new pad/`getFullYear`-based implementation. Leave the `ensureAbsoluteLinkHref` describe block untouched. Verify with `cd client && node --test test/*.test.mjs`.

</details>

Claudia-Anthropica (changes requested):
@bassner Re-reviewing after the develop merge. The date normalization itself is unchanged from what I approved and still looks great, and CodeRabbit's earlier points are all addressed. The blocker is purely a merge artifact: the merge pulled in the #966 guard test (`client/test/format.test.mjs`) whose `formatDate — issue #827` block pins the old `dateStyle: 'medium'` implementation and now reds `run-tests / build` — see inline. Minor aside: `ApplicationListItem.tsx` ('Submission Date', `createdAt`) and `ThesisPage.tsx` (`anonymizedAt`) weren't touched, so they silently dropped their time when `formatDate`'s default flipped from `withTime: true` to date-only — the former now disagrees with `ApplicationData.tsx`, which shows the same field with time. Worth a quick look if that divergence isn't intended.

PR #955, opened 2026-04-28
Add passkey login method

ITegs:
## Overview
Following the docs of the extension plugin this PR adds the client logic and small UI changes to enable redirectless passkeys. 

## Key Features & Changes
* **Authentication Flow & Keycloak**: 
  * Replaced the standard redirect-based Keycloak login with a new Login Modal that offers both "Passkey" and normal password login.
  * Integrated the custom Keycloak provider to expose passkey endpoints (`challenge`, `save`, `authenticate`).
  * Added `silent-check-sso.html` to support Keycloak's silent auth checks.
* **User Interface**:
  * **Login Modal**: The user can decide on passkey or password login.
  * **Header**: Now includes a "Passkey" button for passkey login and "Login" button for password login.
  * **Passkey Registration Prompt**: Automatically suggests eligible users to register a passkey. Users can choose "Never ask again" or "Maybe later".
  * **Account Settings**: Added a new Passkeys settings section where users can register new passkeys, list and delete existing ones.
* **Configuration**:
  * Added `PASSKEY_RP_ID` and `PASSKEY_RP_NAME` to the environment variables (`.env`, Docker configurations, and documentation) to properly configure WebAuthn relying party details.
* **Testing**:
  * Introduced E2E tests (`passkey.spec.ts`) utilizing Playwright's virtual authenticator CDP session to test passkey prompts, settings management, and login flows.
  * Adapted existing auth and data-export tests to handle the new login modal instead of Keycloak page redirects.

## Configuration Details
Two new environment variables have been added:
* `PASSKEY_RP_ID`: The relying party ID (defaults to `window.location.hostname`).
* `PASSKEY_RP_NAME`: The display name shown during registration (defaults to `TUM AET`).
* `PASSKEY_PROMPT_APPS`: Comma separated list of apps shown as supported in the passkey prompt.

## Screenshots
image
image
image
image

> [!NOTE]
> This PR includes out-of-scope changes: DocumentEditor.tsx & EmailTextEditor.tsx
> which fixes failing tests.

## Summary by CodeRabbit

* **New Features**
  * Passkey sign-in in the login modal with improved modal behavior and support detection
  * Global passkey registration prompt with per-user “Never ask again”
  * Passkey management in Settings (list, register, delete) and success/error toasts
  * Silent SSO check endpoint for embedding/status checks
  * Runtime/config option for PASSKEY_RP_ID and platform passkey enablement

* **Tests**
  * End-to-end tests covering passkey registration, login, prompt behavior, persistence, listing, and deletion

* **Documentation**
  * Added PASSKEY_RP_ID configuration guidance and production notes

* **Chores**
  * Keycloak provider integration for passkey support and container updates


docker-compose.yml
```diff
@@ -24,6 +24,7 @@ services:
       KC_BOOTSTRAP_ADMIN_PASSWORD: admin
     volumes:
       - ./keycloak/thesis-management-realm.json:/opt/keycloak/data/import/thesis-management-realm.json:ro
+      - ./keycloak/providers:/opt/keycloak/providers:ro
```
Claudia-Anthropica: @ITegs [high] You're checking in `keycloak/providers/custom-passkey-v1.1.1.jar` (binary, 28KB) and mounting it into the Keycloak container, but there's no upstream source link, no license, and no instructions for rebuilding/upgrading it. That's a supply-chain footgun: future maintainers can't verify what's in the JAR, can't patch CVEs, and can't bump versions without you. Add a short `keycloak/providers/README.md` with the upstream repo URL, the exact release/commit the JAR was built from, the SHA-256 of the file, and the license — and reference it in this PR's description.

In `keycloak/providers/`, the file `custom-passkey-v1.1.1.jar` is a binary blob with no provenance metadata. Add a `keycloak/providers/README.md` that documents: the upstream source repository URL of the custom-passkey extension, the exact tag/commit `v1.1.1` was built from, build instructions, the SHA-256 of the checked-in JAR, and the license under which it's distributed.

</details>

client/src/providers/AuthenticationContext/AuthenticationProvider.tsx
```diff
@@ -8,18 +8,236 @@ import {
...
+      seenCredentialIds.add(normalizedCredential.id)
+      credentials.push(normalizedCredential)
+    }
+  }
+
+  return credentials
+}
+
+const getKeycloakInitOptions = (tokens?: IAuthenticationTokens) => ({
+  onLoad: 'check-sso' as const,
```
Claudia-Anthropica: @ITegs [medium] Adding `onLoad: 'check-sso'` here is a behavior change for *every* user, not just passkey users. Every page load now spawns a hidden iframe to `silent-check-sso.html` and may reconcile the in-memory token with whatever Keycloak session cookie is present. For users who logged in via the password flow and have stored tokens, that silent check can race with `getAuthenticationTokens()` and overwrite the stored tokens (or clear them if the SSO check returns no session). Worth either gating this on the passkey flow only or calling it out explicitly in the PR description and verifying the password flow regression-tests still pass end to end.

In `client/src/providers/AuthenticationContext/AuthenticationProvider.tsx`, `getKeycloakInitOptions` unconditionally sets `onLoad: 'check-sso'`, which changes auth behavior for non-passkey users. Either pass an `onLoad` argument that defaults to omitted for the initial app boot and only `check-sso` when called from `loginWithPasskey`, or add a passkey-flow flag to `initializeKeycloakSession` that toggles the option.

</details>

client/src/pages/SettingsPage/components/PasskeySettings/PasskeySettings.tsx
```diff
@@ -0,0 +1,152 @@
...
+  if (error instanceof DOMException && error.name === 'NotAllowedError') {
+    return 'Passkey request was cancelled or timed out'
+  }
+
+  if (error instanceof Error && error.message.trim().length > 0) {
+    return error.message
+  }
+
+  return 'Passkey operation failed'
+}
```
Claudia-Anthropica: @ITegs [medium] This `getErrorMessage` is the third near-identical copy of the same helper: `getPasskeyErrorMessage` already lives in `client/src/utils/passkey.ts`, and `normalizePasskeyError` in `AuthenticationProvider.tsx` does the same thing. Drop this local copy and the one in the provider, import from `utils/passkey.ts`. Same goes for `isPasskeyCredential` on line 19 — it's identical to the one in `PasskeyRegistrationPrompt.tsx`; promote it to a shared util.

In `client/src/pages/SettingsPage/components/PasskeySettings/PasskeySettings.tsx`, delete the local `getErrorMessage` and import `getPasskeyErrorMessage` from `client/src/utils/passkey.ts` instead. Also delete `isPasskeyCredential` here and the duplicate in `client/src/components/PasskeyRegistrationPrompt/PasskeyRegistrationPrompt.tsx`; export a single `isPasskeyCredential` from `client/src/utils/passkey.ts` and import it in both places. Then remove `normalizePasskeyError` in `client/src/providers/AuthenticationContext/AuthenticationProvider.tsx` and use the shared `getPasskeyErrorMessage` there too.

</details>

client/src/pages/SettingsPage/components/PasskeySettings/PasskeySettings.tsx
```diff
@@ -0,0 +1,152 @@
...
+    try {
+      const allCredentials = await auth.listCredentials()
+      const passkeys = allCredentials
+        .filter(isPasskeyCredential)
+        .sort((a, b) => (b.createdDate ?? 0) - (a.createdDate ?? 0))
+      setCredentials(passkeys)
+    } catch (error) {
+      showSimpleError(getErrorMessage(error))
+    }
+  }
```
Claudia-Anthropica: @ITegs [medium] `refreshCredentials` and the `useEffect` body above (lines 39-52) are the same fetch-filter-sort routine copy-pasted twice. Pull it into a single function and call it from both places — the only real difference is that the initial fetch toggles `isLoadingCredentials`.

In `client/src/pages/SettingsPage/components/PasskeySettings/PasskeySettings.tsx`, extract the credential-fetch logic into a single helper (e.g. `loadPasskeyCredentials`) that calls `auth.listCredentials()`, filters by `isPasskeyCredential`, and sorts by `createdDate` desc. Use it in the `useEffect` (wrapped to toggle `isLoadingCredentials`) and replace `refreshCredentials` with a direct call to the helper.

</details>

client/src/components/PasskeyRegistrationPrompt/PasskeyRegistrationPrompt.tsx
```diff
@@ -0,0 +1,145 @@
...
+        if (isMounted) {
+          setCheckedUserId(userId)
+        }
+
+        if (!hasPasskey && isMounted) {
+          setNeverAskAgain(false)
+          setIsOpen(true)
+        }
+      } catch (error) {
+        console.error('Failed to fetch passkey credentials for login prompt', error)
```
Claudia-Anthropica, line 145: @ITegs [medium] If `auth.listCredentials()` rejects (transient network blip, 401 during a token rotation, etc.), the catch block only logs and `checkedUserId` is never set. Because the effect's dependency array changes on every auth-related re-render, you'll re-fire `listCredentials` over and over for the same user. Set `checkedUserId` (or a separate `attemptedUserId`) in the catch as well so a failure is treated as "checked, with no decision" and stops the loop. Or back off with a flag.

In `client/src/components/PasskeyRegistrationPrompt/PasskeyRegistrationPrompt.tsx`, the catch branch in `checkForPasskeys` only logs and never sets `checkedUserId`, so the effect re-runs the fetch on every subsequent re-render. Add `if (isMounted) setCheckedUserId(userId)` inside the catch block so a single failed attempt is recorded as checked, preventing repeated retries for the same user.

</details>

client/src/providers/AuthenticationContext/AuthenticationProvider.tsx
```diff
@@ -240,6 +479,204 @@ const AuthenticationProvider = (props: PropsWithChildren) => {
...
+          const response = await fetch(getPasskeyEndpoint('authenticate'), {
+            method: 'POST',
+            credentials: 'include',
+            headers: {
+              Accept: 'application/json',
+              'Content-Type': 'application/json',
+            },
+            body: JSON.stringify({
+              credentialId: toBase64Url(passkeyCredential.rawId),
+              rawId: toBase64Url(passkeyCredential.rawId),
```
Claudia-Anthropica: @ITegs [nit] `credentialId` and `rawId` are the same base64url(`rawId`) value here (and in `registerPasskey` at line 603-604). Send one — pick whichever your custom Keycloak extension actually expects and drop the other. Same body, half the bytes, less ambiguity for whoever has to debug this against the extension later.

In `client/src/providers/AuthenticationContext/AuthenticationProvider.tsx`, both `loginWithPasskey` (around line 511-518) and `registerPasskey` (around line 602-608) include `credentialId` and `rawId` set to the same `toBase64Url(passkeyCredential.rawId)`. Verify which field name the custom-passkey Keycloak extension reads and remove the unused one from both request bodies.

</details>

client/src/components/Header/Header.tsx
```diff
@@ -27,9 +33,29 @@ const Header = ({ opened, toggle, authenticatedArea }: HeaderProps) => {
...
+    setIsLoginModalOpen(false)
+    void context.login()
+  }
+
+  const onPasskeyLogin = async () => {
+    setIsPasskeyLoading(true)
+    try {
+      await context.loginWithPasskey()
+      setIsLoginModalOpen(false)
+      navigate('/dashboard')
```
Claudia-Anthropica: @ITegs [nit] `navigate('/dashboard')` runs immediately after `loginWithPasskey()` resolves, but the auth state inside the provider only updates on the next render (state setters are async). It works in practice because the route guard re-evaluates after the state flush, but if you ever land on a guarded route that redirects synchronously on a stale `isAuthenticated`, this will flicker to the login screen and back. Consider awaiting `auth.isAuthenticated` becoming true (or just letting the provider drive the redirect after login).

In `client/src/components/Header/Header.tsx`, `onPasskeyLogin` calls `navigate('/dashboard')` right after `await context.loginWithPasskey()`. Because the auth context state setter inside `loginWithPasskey` is async, the dashboard route may briefly render with `isAuthenticated === false`. Either drop the explicit `navigate` and let the existing route logic redirect when auth state updates, or move the navigation into a `useEffect` that fires when `isAuthenticated` flips to true.

</details>

Claudia-Anthropica (changes requested):
@ITegs Solid first pass on passkey support. The auth provider rework, login modal, settings UI, and e2e coverage are all in good shape. A few things worth tightening before merge: the provenance/license of the binary JAR (this is the biggest one), three near-duplicate helper functions across the new files, a duplicated fetch routine in `PasskeySettings`, and an error path in `PasskeyRegistrationPrompt` that retries forever on transient failures. Also note `getKeycloakInitOptions` now turns on `check-sso` for the whole app, not just passkey users — flagging that as a behavior change. CodeQL/build are green; the e2e job timed out, looks unrelated to the new passkey specs.

Claudia-Anthropica (approved):
@ITegs All previous feedback addressed cleanly — provider README with upstream/SHA/license, `check-sso` gated to the passkey flow, helpers consolidated into `utils/passkey.ts`, the catch branch in `PasskeyRegistrationPrompt` now records `checkedUserId`, the duplicate `rawId` field is gone, and the explicit `navigate('/dashboard')` after passkey login was dropped. Nice work.

client/e2e/passkey.spec.ts
```diff
@@ -0,0 +1,414 @@
...
+          ).credentials.length,
+        { timeout: 30_000 },
+      )
+      .toBeGreaterThan(0)
+  } catch (error) {
+    const message = error instanceof Error ? error.message : String(error)
+
+    // The virtual authenticator sometimes flakes and never reports credentials.
+    // When that happens, we skip the test rather than failing the whole suite.
+    if (message.includes('expect.poll') || message.includes('Timed out')) {
```
Claudia-Anthropica: @ITegs [high] CI is failing right here on both `Passkey - Login` and `Passkey - Registration` — the catch never matches and the original assertion is rethrown. Playwright actually reports `"Timeout 30000ms exceeded while waiting on the predicate"` (not `"Timed out"`), so `message.includes('Timed out')` is false and `'expect.poll'` doesn't appear in the message either. Even if the matcher worked, silently skipping on a poll timeout hides real product bugs — if registration legitimately stops adding the credential, you'd never know. Please diagnose why the virtual authenticator never reports a credential in CI (likely the `/passkey/{client}/challenge` call failing, or an `rp.id`/origin mismatch in the runner) and remove the skip rather than masking it. If you really need a flake escape hatch, mark it `test.fixme()` explicitly.

In `client/e2e/passkey.spec.ts`, the `waitForVirtualPasskey` helper's catch block at lines 51-61 checks `message.includes('expect.poll') || message.includes('Timed out')`, but Playwright's `expect.poll` timeout error message is `"Timeout 30000ms exceeded while waiting on the predicate"`, so neither substring matches and the original assertion is rethrown — this is why the two passkey tests fail in CI. Investigate why `navigator.credentials.create` (or the preceding `/passkey/{client}/challenge` request) doesn't yield a credential on the GitHub Actions runner — most likely the custom Keycloak passkey extension rejects the challenge request or the `rp.id` (set to `window.location.hostname` in `AuthenticationProvider.tsx` `registerPasskey`) doesn't match the runner's origin. Fix the underlying registration failure and remove the entire try/catch silent-skip workaround; do not paper over flakes with auto-skip.

</details>

Claudia-Anthropica (changes requested):
@ITegs The latest commit's CI run is failing on the two new passkey tests — see the inline note. The "Fix CI not capable of passkey registration" workaround doesn't actually catch the timeout because the matcher strings don't match Playwright's error format, so the original assertion is rethrown and the suite fails. Once that's sorted (or the real registration issue diagnosed), this should be ready.

client/e2e/passkey.spec.ts
Claudia-Anthropica: @ITegs Nice — pulling the catch-and-skip and bind-mounting `keycloak/providers` into the CI Keycloak container is the right call. The new passkey tests pass cleanly now. Resolving.

Claudia-Anthropica (approved):
@ITegs All previous feedback addressed cleanly — the catch-and-skip workaround is gone and the CI Keycloak container now bind-mounts `keycloak/providers/`, so the passkey tests actually exercise the registration flow now. The 3 e2e failures in the latest run (`presentation-management.spec.ts:193`, `public-api.spec.ts:17`, `thesis-lifecycle-workflow.spec.ts:103`) are unrelated to this PR — `public-api.spec.ts:17` is already documented flaky on develop, and the other two are generic heading-not-visible timeouts in tests this PR doesn't touch. Worth keeping an eye on, but not blockers. Approving.

client/src/providers/AuthenticationContext/AuthenticationProvider.tsx
```diff
@@ -232,12 +455,215 @@ const AuthenticationProvider = (props: PropsWithChildren) => {
...
+          const userHandle =
+            decodedToken?.sub || decodedToken?.preferred_username || decodedToken?.email || 'user'
+          const username = decodedToken?.preferred_username || decodedToken?.email || userHandle
+          const displayName = decodedToken?.name || username
+          const userIdBytes = new TextEncoder().encode(userHandle).slice(0, 64)
+
+          const passkeyCredential = await navigator.credentials.create({
+            publicKey: {
+              challenge: fromBase64Url(challenge),
+              rp: { name: 'TUM AET', id: GLOBAL_CONFIG.passkey_rp_id },
```
Claudia-Anthropica: @ITegs [medium] You added `PASSKEY_RP_ID` to make the relying party id deployable, but in the same commit hardcoded `rp.name` to `'TUM AET'` — it used to be `GLOBAL_CONFIG.title` (configurable via `APPLICATION_TITLE`/`CHAIR_NAME`). The name shows up in the OS/browser authenticator prompt during registration, so any non-TUM deployment of this image will display "TUM AET" to its users. Either revert to `GLOBAL_CONFIG.title` or pair the id env var with a `PASSKEY_RP_NAME`.

In `client/src/providers/AuthenticationContext/AuthenticationProvider.tsx` at line 563, `rp.name` is hardcoded to the literal `'TUM AET'`. Restore the original tenant-configurable behavior by replacing it with `GLOBAL_CONFIG.title` (or, if a separate brand name is needed, introduce a `PASSKEY_RP_NAME` env var paired with `PASSKEY_RP_ID` and use it here).

</details>

Claudia-Anthropica (changes requested):
@ITegs Mostly polish since my last pass — the new `PASSKEY_RP_ID` env var and label changes look good. One regression worth fixing: the same commit also hardcodes `rp.name` to `'TUM AET'` even though it used to be `GLOBAL_CONFIG.title`, so non-TUM deployments will now show "TUM AET" in their users' authenticator prompt. See inline.

client/src/providers/AuthenticationContext/AuthenticationProvider.tsx
Claudia-Anthropica: @ITegs Friendly bump — looks like the build-fix commit landed but `rp.name` is still hardcoded to `'TUM AET'` at line 563. Mind swapping it back to `GLOBAL_CONFIG.title` (or adding a `PASSKEY_RP_NAME` env to pair with `PASSKEY_RP_ID`)?

Claudia-Anthropica (changes requested):
@ITegs The build-fix and develop merge look fine, but the `rp.name` hardcode I flagged last round is still in place at `AuthenticationProvider.tsx:563`. Once that's swapped back to `GLOBAL_CONFIG.title` (or paired with a `PASSKEY_RP_NAME` env), this should be ready.

Claudia-Anthropica (changes requested):
@ITegs The PASSKEY_RP_ID workflow wiring looks good. Still re-requesting changes because the open feedback from the previous round isn't addressed yet — `rp.name` is still hardcoded to `'TUM AET'` at AuthenticationProvider.tsx:563, and the coderabbitai thread on the challenge fetch (cache: 'no-store') plus the leftover `console.log`s in the keycloak init are also still open. Once those are sorted I think this is good to go.

client/src/providers/AuthenticationContext/AuthenticationProvider.tsx
Claudia-Anthropica: @ITegs Thanks for wiring `PASSKEY_RP_ID` into the deploy workflow — that part's done. Just `rp.name` on this line still needs swapping back from `'TUM AET'` to `GLOBAL_CONFIG.title` (or a paired `PASSKEY_RP_NAME` env). It's a one-line change.

Claudia-Anthropica (changes requested):
@ITegs The deploy workflow now wires `PASSKEY_RP_ID` through correctly — that bit's good. Re-requesting changes only because the open thread on `AuthenticationProvider.tsx:563` is still unresolved: `rp.name` is hardcoded to `'TUM AET'` and needs to flow from `GLOBAL_CONFIG.title` (or a new `PASSKEY_RP_NAME` env). One-line fix and you're done.

Claudia-Anthropica (changes requested):
@ITegs Nice update on the registration prompt copy — much friendlier. Re-requesting changes only because the two open threads from the previous round are still unresolved on the current HEAD: `rp.name` is still hardcoded to `'TUM AET'` at AuthenticationProvider.tsx:568 (needs `GLOBAL_CONFIG.title` or a paired `PASSKEY_RP_NAME`), and coderabbit's thread on the challenge fetch (`cache: 'no-store'` + the leftover `console.log` calls at lines 311/315/318) is still open too. Both are one-line/few-line fixes.

client/src/providers/AuthenticationContext/AuthenticationProvider.tsx
Claudia-Anthropica: @ITegs Bumping again — line 601 still reads `rp: { name: 'TUM AET', id: GLOBAL_CONFIG.passkey_rp_id }`. The id is configurable now (good), but the name is still hardcoded. One-line fix — please either swap to `GLOBAL_CONFIG.title` or add a paired `PASSKEY_RP_NAME` env. Happy to push it for you if it's easier.

client/src/app/layout/AuthenticatedArea/AuthenticatedArea.tsx
```diff
@@ -175,6 +164,27 @@ const AuthenticatedArea = (props: PropsWithChildren<IAuthenticatedAreaProps>) =>
...
+      hasTriggeredFallbackLogin.current = false
+      return
+    }
+
+    if (location.pathname === '/logout' || hasTriggeredFallbackLogin.current) {
+      return
+    }
+
+    hasTriggeredFallbackLogin.current = true
+    void login(`${location.pathname}${location.search}${location.hash}`)
```
Claudia-Anthropica: @ITegs [high] The new `validatePasskeyAvailability` check is async, but `isPasskeyExtensionAvailable` initial state is `false` — which means `auth.isPasskeySupported` starts `false` for everyone on first render, then flips to `true` once the fetch resolves. This effect fires on that first render and redirects to Keycloak before the check completes, so any unauthenticated user who navigates directly to a protected route (bookmark, refresh, deep link) bypasses the passkey login modal entirely — even on a Keycloak that has the extension. The passkey path only really works coming from the public landing page.

Easiest fix: make `isPasskeyExtensionAvailable` tri-state (e.g. initial `undefined`, set to `true`/`false` after the fetch), and skip the fallback while it's still `undefined`.

In `client/src/app/layout/AuthenticatedArea/AuthenticatedArea.tsx` lines 167-186, the fallback effect calls `login()` whenever `isPasskeySupported` is false, but `isPasskeySupported` derives from `isPasskeyExtensionAvailable` which starts at `false` in `AuthenticationProvider.tsx` (line 196) and only resolves to its real value after an async fetch. Change `isPasskeyExtensionAvailable` to `useState<boolean | undefined>(undefined)` and expose `isPasskeySupported` on the context as `boolean | undefined`, then in the fallback effect skip the redirect while `isPasskeySupported === undefined`. The fallback must only fire after the availability check has resolved.

</details>

Claudia-Anthropica (changes requested):
@ITegs The passkey-availability gating is a nice idea, but the new fallback effect in `AuthenticatedArea` fires before the availability check resolves, so unauthenticated users hitting a protected route get redirected to Keycloak before they ever see the passkey modal. Also, `rp.name` on line 601 is still hardcoded to `'TUM AET'` — replied on the existing thread for the fourth time, please take a look.

client/src/config/global.ts
```diff
@@ -128,6 +128,8 @@ export const GLOBAL_CONFIG: IGlobalConfig = {
     /\/+$/,
     '',
   ),
+  passkey_rp_id: getEnvironmentVariable('PASSKEY_RP_ID')?.trim() || window.location.hostname,
+  passkey_rp_name: getEnvironmentVariable('PASSKEY_RP_NAME')?.trim() || 'TUM AET',
```
Claudia-Anthropica: @ITegs [high] `PASSKEY_RP_NAME` is consumed by `GLOBAL_CONFIG.passkey_rp_name` here, but the env var isn't wired anywhere downstream — it's missing from the allowlist in `client/public/generate-runtime-env.js`, from `docker-compose.prod.yml`'s `client` environment block, from the envs list (and build-arg) in `.github/workflows/deploy_docker.yml`, and from `docs/CONFIGURATION.md`. So in production it will always fall back to `'TUM AET'` regardless of what operators set — the configurability you just added doesn't actually take effect. Mirror the wiring you already did for `PASSKEY_RP_ID`.

In `client/src/config/global.ts:132`, `passkey_rp_name` reads from `PASSKEY_RP_NAME` but the env var is not exposed at runtime or in deployment. Add `'PASSKEY_RP_NAME'` to the `ALLOWED_ENVIRONMENT_VARIABLES` array in `client/public/generate-runtime-env.js`, add `- PASSKEY_RP_NAME` under the client service's `environment:` block in `docker-compose.prod.yml`, append `,PASSKEY_RP_NAME` to the `envs:` list and add `PASSKEY_RP_NAME: ${{ vars.PASSKEY_RP_NAME }}` plus a `PASSKEY_RP_NAME=${PASSKEY_RP_NAME}` build-arg line in `.github/workflows/deploy_docker.yml`, and add a `PASSKEY_RP_NAME` row to the table in `docs/CONFIGURATION.md` next to `PASSKEY_RP_ID`.

</details>

keycloak/providers/README.md
```diff
@@ -0,0 +1,46 @@
+# Keycloak Provider Artifacts
+
+This directory contains checked-in Keycloak provider JARs that are copied into the Keycloak container at startup.
+
+## custom-passkey-v1.2.0.jar
```
Claudia-Anthropica: @ITegs [medium] You bumped the jar to `custom-passkey-v1.3.0.jar` in this commit, but the README still describes v1.2.0 throughout — version heading, upstream tag, embedded coordinates, release-tag URL, install filename, `git checkout` step, and the SHA-256 checksum all point at the old version. Anyone following these instructions will end up with the wrong jar. Update every occurrence and recompute the SHA-256 of the v1.3.0 binary.

In `keycloak/providers/README.md`, every section still describes `custom-passkey-v1.2.0.jar` but the actual checked-in artifact is `custom-passkey-v1.3.0.jar`. Replace every occurrence of `v1.2.0` / `1.2.0` with `v1.3.0` / `1.3.0`, update the upstream commit SHA to whatever commit the v1.3.0 tag points at on https://github.com/ITegs/keycloak-redirectless-passkey, and replace the SHA-256 line with `80f291e745cb28643e156014a50d3eb6178c117f57bdeee3baefcb48c358221f  custom-passkey-v1.3.0.jar` (the actual digest of the checked-in binary).

</details>

client/e2e/passkey.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+    } finally {
+      await teardownVirtualAuthenticator(cdpSession, authenticatorId)
+    }
+  })
+})
+
+test.describe('Passkey - Login', () => {
+  test.use({ storageState: authStatePath('examiner2') })
+
+  test('signs in with a registered passkey from the login modal', async ({ page }) => {
```
Claudia-Anthropica, line 256: @ITegs [high] This test is failing on the current HEAD in CI: after `loginModal.getByRole('button', { name: 'AET Passkey' }).click()` the URL becomes `/dashboard` (line 235 passes) but `getByRole('heading', { name: /dashboard/i })` never appears within 30s on line 236, and the retry attempts can't even reopen `/settings/account` to find the `Passkeys` heading. The 7 other examiner/examiner2 failures in the same run look like cascading flakiness from the new login-modal-based `auth.setup.ts` flow, but this passkey login failure is specifically tied to the auth flow this PR introduces. Please reproduce locally and figure out what's blocking the dashboard render after a passkey login.

In `client/e2e/passkey.spec.ts:218` (`signs in with a registered passkey from the login modal`), the test fails: after the passkey-login click `expect(page).toHaveURL(/\/dashboard/)` passes but the dashboard heading never becomes visible. Reproduce with `cd client && npx playwright test e2e/passkey.spec.ts --grep "registered passkey from the login modal"` against a local stack (`docker compose up -d` for Keycloak+Postgres, then `cd server && ./gradlew bootRun --args='--spring.profiles.active=dev'` and `cd client && npm run dev`). Investigate whether `AuthenticationProvider` correctly transitions `isAuthenticated`/`user` after the passkey token exchange and whether `AuthenticatedArea` advances past `<PageLoader />` (the `auth.user ? ... : <PageLoader />` branch) once the user has loaded. Fix the regression so the dashboard renders post-passkey-login and the test passes.

</details>

Claudia-Anthropica (changes requested):
@ITegs Mostly cleanup since last round — `rp.name` is now configurable via `passkey_rp_name`, and the `AuthenticatedArea` race is gone. Two new issues from `Fix claudia issues`: the `PASSKEY_RP_NAME` env var isn't wired through to runtime/deploy, so the configurability you added doesn't actually take effect, and the providers README still describes `v1.2.0` while the binary is `v1.3.0`. Also the e2e passkey login test is red on the current HEAD — see inline.

ITegs:
@Claudia-Anthropica review

Claudia-Anthropica:
@ITegs Error 403: Insufficient permissions to send me on missions. Maybe ask a maintainer?

keycloak/providers/custom-passkey-v1.3.1.jar
krusche, line 1: we should not add jars in the git repository, instead this should be a new gradle dependency

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

krusche (changes requested):
Here is my feedback:
* Please provide a better PR description
* It looks like tests are missing, in particular e2e tests would be important to make sure the functionality works in the future, but also client and server tests would be good
* The jar file solution is bad, please reference the dependency using gradle

ITegs:
> * It looks like tests are missing, in particular e2e tests would be important to make sure the functionality works in the future, but also client and server tests would be good

e2e tests for the passkey actions are included in `passkey.spec.ts`. The server is independent of the passkey implementation; it only happens between Keycloak and the client. That's why there are no server tests.

krusche (changes requested):
## Deep Review — PR #955 (Passkey login)

Thanks for the thorough work — the WebAuthn flows, virtual-authenticator E2E tests, and the per-user "Never ask again" affordance are all well thought through. A few things I would like addressed before merge, plus several smaller suggestions.

### 🔴 Critical / must address before merge

**1. Operational gap: production Keycloak rollout is not documented**
The PR mounts `keycloak/providers` into the dev Keycloak container and sets `webAuthnPolicyPasswordlessPasskeysEnabled: true` in `keycloak/thesis-management-realm.json`, but per the project convention these are *dev-only* artifacts (see `CLAUDE.md` → "Keycloak Configuration (Dev Only)"). Production runs a separate Keycloak instance. The only production-facing doc change is adding `PASSKEY_RP_ID=` to `docs/PRODUCTION.md`. Without an explicit production runbook the deployment will silently appear to "work" (login modal shows the "AET Passkey" button) until a user actually presses it and gets a 404 against `realms/<realm>/passkey/<clientId>/challenge`.

Please add a section to `docs/PRODUCTION.md` covering:
- installing `custom-passkey-v1.3.1.jar` into the prod Keycloak `providers/` directory and restarting (or rebuilding) Keycloak so it gets loaded
- enabling `webAuthnPolicyPasswordlessPasskeysEnabled` (and aligning `webAuthnPolicyPasswordless*` knobs) on the production realm
- setting `PASSKEY_RP_NAME` (`docs/PRODUCTION.md` only mentions `PASSKEY_RP_ID`, but the docker-compose, generate-runtime-env, and the deploy workflow all need both — this is inconsistent across the PR)
- the requirement that `PASSKEY_RP_ID` matches Keycloak's WebAuthn Passwordless RP ID exactly

**2. Supply-chain provenance for the bundled Keycloak provider**
`keycloak/providers/custom-passkey-v1.3.1.jar` is a 30KB binary checked into the repo, sourced from `ITegs/keycloak-redirectless-passkey` (the PR author's personal account). The README documents the SHA-256 (good) but:
- there is no release signature or attestation
- the SHA is documented *next to* the binary, so an attacker compromising the repo would update both
- in production this JAR will be loaded into Keycloak with full SPI privileges

At a minimum, please document who is responsible for reviewing future bumps. Ideally:
- consider publishing the provider via Maven Central or a TUM-controlled artifact registry with checksum/sig verification on pull, instead of checking in a binary, or
- move the JAR under a sub-tree owned by the chair's `aet`/`ls1intum` GitHub org, or
- have the build pipeline rebuild from source (the README documents the steps) and pin the upstream commit `854f6ef1d3e43947c186bdd789841af4924ec623` in CI.

**3. `Header.tsx`: `opened={isLoginModalForcedOpen ?? isLoginModalOpen}` — `??` is the wrong operator**

```tsx
const isLoginModalForcedOpen =
  context.isReady && openLoginModal && !context.isAuthenticated && context.isPasskeySupported
...
<Modal opened={isLoginModalForcedOpen ?? isLoginModalOpen} ...
```

`isLoginModalForcedOpen` is a boolean (or `undefined` only while `isPasskeySupported` is being probed). When it resolves to `false`, `?? ` does **not** fall through — so the modal will never honor `isLoginModalOpen` after the probe completes. Today this happens to be harmless (because the `useEffect` that sets `isLoginModalOpen` itself requires the same conditions that would force the modal open), but it is fragile and confusing; the next change to either condition will introduce a real bug. Use `||` and drop the duplicate condition tree, or compute one value:

```tsx
const isLoginModalOpen = context.isReady && openLoginModal && !context.isAuthenticated && context.isPasskeySupported
```

…and remove the parallel `useState` entirely (the modal can be opened from the header button by a separate signal; right now `setIsLoginModalOpen(true)` is dead code on the forced-open path).

**4. `registerPasskey`: WebAuthn `user.id` derivation is unsafe**

```ts
const userHandle =
  decodedToken?.sub ?? decodedToken?.preferred_username ?? decodedToken?.email ?? 'user'
...
const userIdBytes = new TextEncoder().encode(userHandle).slice(0, 64)
```

Two problems:

a) Fallback chain. `sub` is the Keycloak user UUID; `preferred_username`/`email` are mutable. If a user ever registers a passkey while `sub` is missing for any reason, then registers later with `sub` present, they will have *two separate* WebAuthn user handles for the same Keycloak account — leading to passkeys that authenticate to the wrong subject (or are silently overwritten by Keycloak). The right behavior is to require `sub` and throw if absent.

b) `.slice(0, 64)` on the byte array can split a multi-byte UTF-8 character mid-stream. For `sub` (always ASCII UUID, 36 bytes) this never triggers; but with the email fallback it can yield malformed bytes that will fail to match anything on subsequent logins. Either drop the fallback (recommended), or right-pad/truncate by codepoint, not by byte.

Suggested fix:
```ts
const userHandle = decodedToken?.sub
if (!userHandle) {
  throw new Error('Cannot register a passkey: access token has no subject')
}
const userIdBytes = new TextEncoder().encode(userHandle) // sub is a UUID, well under 64 bytes
```

**5. `AuthenticatedArea.tsx`: fallback-to-password-login latch can leave the user stuck**

```ts
hasTriggeredFallbackLogin.current = true
void login(`${location.pathname}${location.search}${location.hash}`)
```

If the Keycloak redirect fails (network error, popup blocker, …) the ref is never reset on this branch — only on `isAuthenticated || isPasskeySupported` becoming true. A user who lands back on the route still unauthenticated will see the page hang with no login affordance. Either:
- reset the ref in the `.catch` of `login(...)` (currently `login` returns `unknown`, so the type makes this awkward — consider tightening the return to `Promise<void>`), or
- gate the latch on `location.pathname` so route changes re-arm the fallback.

### 🟠 Important

**6. `PasskeyRegistrationPrompt` re-fetches credentials whenever the `auth` context identity changes**
The effect has `auth` in its dependency array; the context value is recreated by `useMemo` in `AuthenticationProvider` whenever `user`, `access_token`, etc. change. The `checkedUserId === userId` guard prevents re-opening the modal, but the API call to `listCredentials()` (which itself can trigger a Keycloak token refresh) still fires every time the guard *evaluates*, until `setCheckedUserId(userId)` lands. Drop `auth` from the deps and pull the specific functions/values you need:

```ts
}, [auth.isAuthenticated, auth.isPasskeySupported, auth.listCredentials, checkedUserId, location.pathname, shouldSkipPrompt, userId])
```

Also: `shouldSkipPrompt` is a `useMemo` that reads `localStorage` but its dependency keys only change when `userId` changes. After the user clicks "Maybe later", `persistPromptDismissal` writes to `localStorage` but `shouldSkipPrompt` is stale until the next userId change. Right now this is masked because `setIsOpen(false)` keeps the modal closed, but the next code that reads `shouldSkipPrompt` for any other purpose will see stale data. Either read `localStorage` inline at the point you need it, or store the dismissal in component state in addition to `localStorage`.

**7. The login modal is the *only* unauthenticated entry point — and it is gated on `context.isPasskeySupported`**
Per `Header.tsx` lines 173-189 and `AuthenticatedArea.tsx` lines 167-190, when `isPasskeySupported` is `false`, unauthenticated users on a protected route are redirected straight to Keycloak (good). But when `isPasskeySupported` is `undefined` (the brief window during the `/passkey/health` probe), both the modal and the redirect are blocked — the user sees a blank page with nothing happening. If the health-check endpoint hangs (slow network, Keycloak unreachable) the experience is opaque.

Suggestions:
- short-circuit `isPasskeyExtensionAvailable` to `false` after a reasonable timeout (e.g. 3s) so the user is never stuck in `undefined`
- show a `<PageLoader />` while the probe is in flight on protected routes
- log a warning when the probe fails so production issues are visible in Sentry/console

**8. Out-of-scope changes bundled into the PR**
The `Fix loading errors in RichTextEditor` commit modifies `DocumentEditor.tsx` and `EmailTextEditor.tsx` (≈150 lines) which are unrelated to passkey login. These changes look reasonable in isolation (guarding against destroyed-editor races after the tiptap 3.23 schema-null behavior), but they should be a separate PR — they have nothing to do with passkeys and shouldn't be reviewed together. At a minimum, please explain in the PR description why they are here so reviewers know what they are signing off on.

**9. Hardcoded "AET-only" content in the registration prompt**
```ts
const AET_CHAIR_URL = 'https://aet.cit.tum.de/'
...
{['Thesis Management', 'TUM Apply', 'Prompt'].map((appName) => ...)}
```

This deeply couples a generic auth component to TUM AET. The codebase already has `GLOBAL_CONFIG.chair_name` / `chair_url`. Please source the chair URL from config (defaulting to `chair_url`), and either move the "linked apps" badges behind an env-configurable list or remove them — they are not really truthful (a passkey registered here does not actually let you sign in to TUM Apply unless that deployment shares the same Keycloak RP). False UX promises in an auth dialog are particularly bad.

**10. `PASSKEY_RP_NAME` not in `docs/PRODUCTION.md` example**
The deploy workflow, both docker-compose files, `generate-runtime-env.js`, and `global.ts` all read `PASSKEY_RP_NAME`, but `docs/PRODUCTION.md` only adds `PASSKEY_RP_ID=`. Operators following the docs will see "TUM AET" as the browser passkey prompt name on every deployment regardless of branding. Add `PASSKEY_RP_NAME=` to the example environment.

### 🟡 Suggestions

**11. `getPasskeyErrorMessage` is doing two unrelated jobs**
It both extracts error text from a `Response`/`Error` and applies operation-specific overrides (`register/409` → "already registered"). The 4xx-status → "invalid passkey" mapping for the `login` op is also a bit aggressive — a 401 from a misconfigured server will look like "this passkey is not valid" to the user, which is misleading. Consider narrowing to specific known statuses, or surface the upstream message on unexpected ones.

**12. `getDeviceName` over-defensive casting**
```ts
const navigatorPlatformRaw = (navigator as unknown as Record<string, unknown>).platform
const navigatorPlatform = typeof navigatorPlatformRaw === 'string' ? navigatorPlatformRaw.trim() : ''
```
`navigator.platform` is in the lib.dom types (deprecated but typed as `string`). The double cast is unnecessary — `navigator.platform.trim()` works directly. Similar for `navigator.userAgent`.

**13. `formatCreatedDate` heuristic is brittle**
```ts
const milliseconds = createdDate > 10_000_000_000 ? createdDate : createdDate * 1000
```
Auto-detecting seconds vs. milliseconds by magnitude is fine for now but will quietly break in 2286. Keycloak's `/account/credentials` actually documents milliseconds — pick one and assert it.

**14. Storage-key duplication between component and tests**
`PasskeyRegistrationPrompt.tsx` declares `NEVER_ASK_AGAIN_STORAGE_KEY`, `MAYBE_LATER_STORAGE_KEY`, `DISABLE_PROMPT_STORAGE_KEY`; `passkey.spec.ts` redeclares them as `..._PREFIX` constants. If anyone renames in one place but not the other, the tests will silently stop verifying the right key. Export them from a single module (e.g. `client/src/utils/passkey.ts`) and import in both production and test code.

**15. Re-using the `keycloak` module-level binding**
`request.ts` imports `{ keycloak }` from `AuthenticationProvider`. ESM live bindings make this work correctly even after `keycloak = createKeycloakClient()` re-assignment, but it is a non-obvious dependency that pretty much any "let's tree-shake" or "let's pre-bundle" tooling change could break. Consider exposing a `getKeycloak()` accessor instead — cheap to add now, painful to refactor later.

**16. `closeOnClickOutside={false}` on the login modal**
The login modal cannot be dismissed by clicking outside (only by Escape or close button). This is intentional per the code, but the registration prompt allows click-outside while not registering. Pick a consistent convention.

**17. Test user credentials in clear text in the repo**
`passkey_user/passkey_user` is added to both `seed_dev_test_data.sql` and the realm export. Existing convention in this codebase (e.g. `student/student`) — fine to follow, but if you want to tighten any one of these, this PR is a good place to do it.

**18. `silent-check-sso.html` has no `<title>` and no CSP meta**
Minor — Keycloak's docs use exactly this content, but consider adding `<title>Silent SSO Check</title>` so the page has a name in browser history / accessibility tooling, and a restrictive `<meta http-equiv="Content-Security-Policy" content="default-src 'none'; script-src 'unsafe-inline'">` to harden the iframe target.

### 🟢 Things done well
- The virtual-authenticator E2E coverage (`passkey.spec.ts`) is excellent — it exercises the prompt, settings, login, and persistence flows end-to-end against a real Chrome WebAuthn implementation.
- Defensive handling of the `/passkey/health` probe means the UI degrades gracefully when the provider isn't installed in non-prod Keycloak.
- Per-user `localStorage` keys with a global "disabled" master switch is a clean design for E2E control.
- `parseAccountCredentialsResponse` is appropriately defensive about the two shapes Keycloak returns; the de-duplication via `seenCredentialIds` is a nice touch.
- The `getPasskeyEndpoint`/`getAccountCredentialsEndpoint` helpers URL-encode realm and clientId — easy to overlook.
- Token refresh handling (`isTokenExpired(5)` → `updateToken(5 * 60)`) before each authenticated passkey call is correct.

### Verification I'd like to see before merge
- A short note in the PR description on how the production Keycloak rollout will be sequenced (provider jar install, realm policy flip, env var rollout). This is the highest-risk part of the change and is currently invisible.
- Either remove the AET-specific badges/copy or gate them behind config.
- Address #3 (the `??` operator), #4 (user handle), and #5 (fallback latch) — these are simple fixes and reduce real risk.

Happy to re-review once these are in.

Claudia-Anthropica (approved):
@ITegs All my previous feedback is addressed cleanly — `rp.name` now reads `GLOBAL_CONFIG.passkey_rp_name`, the new `PASSKEY_PROMPT_APPS` env var is wired through all four config layers, the README matches the jar version with SHA + license, and the prompt-retry loop is bounded properly now. Nice cleanup pass, approving from my side. Heads-up: there's still an open thread from @krusche about packaging the jar as a Gradle dependency rather than checking the binary in — worth resolving with him before merge.

krusche (changes requested):
## Re-review — PR #955 (Passkey login)

Thanks for the careful turnaround. Most of the previous critical findings are properly addressed and the Maven-Central provider rewrite in particular is a big improvement. There is one **new regression** introduced by the merge commit that needs fixing, plus a couple of leftovers from the refactor.

### ✅ Now resolved
- **Provider supply chain (was #2)** — `custom-passkey-v1.3.1.jar` removed, replaced by a Gradle pull of `de.tum.cit.aet:keycloak-redirectless-passkey:1.3.2` from Maven Central via a multi-stage `keycloak/Dockerfile`. Dev, showcase, and CI compose files all build the image the same way. Excellent.
- **`registerPasskey` user handle (was #4)** — `decodedToken?.sub` only, with a hard error if missing. ✅
- **Header `??` operator (was #3)** — collapsed to a single `isLoginModalOpen` expression that genuinely drives the modal. ✅
- **PasskeyRegistrationPrompt deps & stale memo (was #6)** — destructured `auth`, moved `listCredentials` into a ref, added a `checkingUserIdRef` re-entrancy guard, and `shouldSkipPrompt` is now state with `setShouldSkipPrompt(true)` in `persistPromptDismissal`. ✅
- **Blank page during passkey probe (was #7)** — new `isWaitingForPasskeyProbe` shows a `<PageLoader />` while `isPasskeySupported === undefined`. ✅
- **Hardcoded AET branding (was #9)** — chair link uses `GLOBAL_CONFIG.chair_name` / `chair_url`; sibling apps are configurable via `PASSKEY_PROMPT_APPS` and the prompt copy adapts when the list is empty. ✅
- **`PASSKEY_RP_NAME` in prod docs (was #10)** — now in `docs/PRODUCTION.md` next to `PASSKEY_RP_ID` and `PASSKEY_PROMPT_APPS`. ✅
- **`silent-check-sso.html` CSP / title (was #18)** — both added. ✅

### 🔴 Critical — new regression introduced by this revision

**The `LIQUIBASE_CONTEXTS` production safety warning was deleted from `docs/CONFIGURATION.md`.**

On `develop`, that row reads:

> | LIQUIBASE_CONTEXTS | server | prod | Comma-separated list of Liquibase contexts to execute on startup. **Must be `prod` (or unset) in production** — the value `dev` activates the `23_seed_dev_test_data.xml` changelog, which seeds fake users, topics, applications, and theses, and `application-dev.yml` already sets this to `dev` for the dev profile. Never set this to `dev` (or include `dev`) on a production deployment. |

On this PR (`docs/CONFIGURATION.md:43`) the description cell is now empty:

```
| LIQUIBASE_CONTEXTS             | server         | prod  |                |
```

The merge commit `86a3a1b4` ("Merge branch 'develop' into feat/add-passkey-login") resolved a column-width conflict on this row by dropping the warning text. This is unrelated to passkeys but it shipped on this PR, and the warning is genuinely security-critical (it is the only place that explains how a production operator can accidentally seed fake admin/student users). Please restore the original description verbatim.

### 🟠 Important

**1. Dead `useEffect` left over in `Header.tsx`**

`Header.tsx:49-58`:

```tsx
useEffect(() => {
  if (
    !context.isReady ||
    !openLoginModal ||
    context.isAuthenticated ||
    !context.isPasskeySupported
  ) {
    return
  }
}, [context.isAuthenticated, context.isPasskeySupported, context.isReady, openLoginModal])
```

Every branch is `return` with no side-effect. This effect does nothing and should be deleted; it's a leftover from the previous `setIsLoginModalOpen` state machine you removed in the same commit.

**2. `passkey_rp_id` / `passkey_rp_name` default fallback uses the wrong operator**

`client/src/config/global.ts:142-143`:

```ts
passkey_rp_id: getEnvironmentVariable('PASSKEY_RP_ID')?.trim() ?? window.location.hostname,
passkey_rp_name: getEnvironmentVariable('PASSKEY_RP_NAME')?.trim() ?? 'Thesis Management',
```

`??` only falls through on `undefined` / `null`, but `.trim()` produces an *empty string* for whitespace-only env values, which is then accepted as the RP ID. A WebAuthn `rpId` of `""` will fail every passkey ceremony with cryptic browser errors. This was `||` in your previous revision (which fell through correctly on `""`). Either revert to `||`:

```ts
passkey_rp_id: getEnvironmentVariable('PASSKEY_RP_ID')?.trim() || window.location.hostname,
```

…or, cleaner, make `getEnvironmentVariable` itself treat whitespace-only values as missing.

(In practice this is unlikely to bite because most ops set values cleanly, but the diff *added* a regression here vs. the prior version — `||` → `??` — so it's worth a one-character fix.)

### 🟠 Still unresolved from the previous review

**3. `AuthenticatedArea.tsx` fallback-login latch (was #5)** — `hasTriggeredFallbackLogin.current = true` is set immediately before `void login(...)`, and is only reset on `isAuthenticated || isPasskeySupported`. If `login()` errors (network blip, popup blocked), the user is stuck with no further attempts. At minimum reset the ref in `.catch`:

```ts
hasTriggeredFallbackLogin.current = true
login(`${location.pathname}${location.search}${location.hash}`).catch((error) => {
  console.error('Fallback login failed', error)
  hasTriggeredFallbackLogin.current = false
})
```

This requires tightening `login`'s return type from `unknown` to `Promise<void>` (it already returns a thenable in the implementation).

**4. Unrelated tiptap fixes still bundled (was #8)** — `DocumentEditor.tsx` and `EmailTextEditor.tsx` are still part of the diff (+103/-45 between them). These are not passkey work and should land in their own PR so reviewers can sign off on them separately. At a minimum, please call them out explicitly in the PR description so they aren't a stealth change.

**5. Production Keycloak realm configuration (partial of #1)** — the new Dockerfile installs the provider automatically, which is a great improvement, but `docs/PRODUCTION.md` still does not tell prod operators that they must enable `webAuthnPolicyPasswordlessPasskeysEnabled` (and align the rest of the `webAuthnPolicyPasswordless*` knobs) on the **production** realm. The dev realm import has it set; per `CLAUDE.md` prod uses a separate Keycloak with its own configuration, so this is a manual step. Worth one sentence in PRODUCTION.md so the first production rollout does not silently 404 on `/passkey/challenge`.

### 🟡 Minor

- `Header.tsx:60-65`: `onLoginModalClose` is now a one-branch if-return; the body simplifies to `if (isLoginModalOpen) void navigate('/', { replace: true })`.
- `Header.tsx:183` `opened={isLoginModalOpen ?? false}` — `isLoginModalOpen` is now typed `boolean | undefined` (because `isPasskeySupported` can be `undefined`); `?? false` here is correct. Just noting the typing chain in case you want to narrow the context type later.
- E2E label tracking is now consistent across `auth.spec.ts`, `passkey.spec.ts` and the component — nice cleanup.

### Recommended action before merge
1. Restore the `LIQUIBASE_CONTEXTS` description in `docs/CONFIGURATION.md` (the regression).
2. Remove the dead `useEffect` in `Header.tsx:49-58`.
3. Revert `??` → `||` (or equivalent fix) for `passkey_rp_id` / `passkey_rp_name`.
4. Decide on the unrelated tiptap commit — split out or call out.
5. Either fix the fallback-login latch or open a follow-up issue and link it from the PR.
6. Add a one-line "enable WebAuthn passwordless / passkeys on the production realm" note to `docs/PRODUCTION.md`.

Items 1–3 are quick; items 4–6 can be tracked as follow-ups if you'd rather not expand scope further. Happy to re-review once #1–3 are in.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

client/src/pages/SettingsPage/components/PasskeySettings/PasskeySettings.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  if (!auth.isPasskeySupported) {
+    return null
+  }
+
+  return (
+    <Stack>
+      <Title order={3}>Passkeys</Title>
+      <Text>Register a passkey on this device to sign in across AET apps.</Text>
```
Claudia-Anthropica: @ITegs [low] This copy still hardcodes "AET apps", but the registration prompt was reworked to be tenant-configurable (chair name + `PASSKEY_PROMPT_APPS`). On a non-AET deployment the settings page reads "sign in across AET apps" while the prompt shows the configured branding — inconsistent. Drop the brand here or derive it the same way the prompt does.

In `client/src/pages/SettingsPage/components/PasskeySettings/PasskeySettings.tsx` line 86, the description text hardcodes "across AET apps". Replace it with tenant-neutral copy such as "Register a passkey on this device for faster, more secure sign-in.", or derive the brand from `GLOBAL_CONFIG.chair_name` to match the configurable copy already used in `PasskeyRegistrationPrompt`.

</details>

Claudia-Anthropica (changes requested):
@ITegs Took another full pass on the current head. Since my earlier approval the develop merge brought in a couple of regressions, and @krusche's review from earlier today still applies on this exact commit — I won't duplicate his inline notes, but I'm aligned with that list. The dropped `LIQUIBASE_CONTEXTS` production warning in `docs/CONFIGURATION.md` is the one that genuinely matters (it's the safeguard that stops an operator from seeding fake admin/student users into a prod DB), so that description should go back verbatim. One small extra thing from me inline; otherwise the passkey work itself still looks solid.

Claudia-Anthropica (approved):
@ITegs All my outstanding feedback is addressed in the "Fix code review issues" commit — the `LIQUIBASE_CONTEXTS` production warning is restored verbatim in `docs/CONFIGURATION.md`, the dead `useEffect` in `Header.tsx` is gone, the `passkey_rp_id`/`passkey_rp_name` fallbacks are back to `||` (so whitespace-only env values no longer slip through), the AuthenticatedArea fallback-login latch now resets in `.catch` with `login` typed as `Promise<void>`, and the WebAuthn-passwordless prod note landed in `PRODUCTION.md`. The develop merge only pulled in the SBOM/dep work and CI is green. Nice work on the turnaround — approving.

client/src/utils/passkey.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  if (
+    (operation === 'login' && status !== undefined && status >= 400 && status < 500) ||
+    message === invalidPasskeyMessage
+  ) {
+    return invalidPasskeyMessage
+  }
+
+  return fallback
+}
```
krusche, line 72: WebAuthn-specific error names aren't handled here. When the user dismisses the OS prompt, `navigator.credentials.get/create` rejects with a `DOMException` whose `name` is `NotAllowedError` (cancellation/timeout) or `AbortError`. Both `instanceof Error` is true, so we fall through to the generic fallback message.

Suggest branching early on `error instanceof DOMException`:
```ts
if (error instanceof DOMException) {
  if (error.name === 'NotAllowedError' || error.name === 'AbortError') {
    return 'Passkey sign-in was cancelled.'
  }
  if (error.name === 'InvalidStateError') {
    return duplicatePasskeyMessage
  }
  if (error.name === 'SecurityError') {
    return 'Passkey is not allowed on this site.'
  }
}
```
And ideally suppress the toast entirely on cancellation (return a sentinel the caller checks).

client/src/providers/AuthenticationContext/AuthenticationProvider.tsx
```diff
@@ -5,135 +5,408 @@ import { AuthenticationContext } from './context'
...
+    const storedTokens = getAuthenticationTokens()
+
+    void initializeKeycloakSession(storedTokens)
       .then(() => {
-        storeTokens()
         triggerReadySignal()
       })
       .catch((error) => {
         console.error('Keycloak init error', error)
       })
```
krusche, line 324: Silent failure: if `keycloak.init` rejects (Keycloak unreachable, CSP blocking the silent-check-sso iframe, malformed realm), this only logs to the console and the `finally` block below still triggers `triggerReadySignal()`. The UI then renders an unauthenticated state with no indication of *why*.

Please surface this via `showSimpleError('Authentication service is unavailable. Please try again.')` so users (and on-call) have a chance to diagnose. The current `console.error` is invisible in production.

client/src/providers/AuthenticationContext/AuthenticationProvider.tsx
```diff
@@ -5,135 +5,408 @@ import { AuthenticationContext } from './context'
...
         }
       })
-    }
-
-    const storeTokens = () => {
-      const refreshToken = keycloak.refreshToken
-      const accessToken = keycloak.token
+      .catch(() => {
+        setAuthenticationTokens(undefined)
+      })
```
krusche, line 212: Bare `.catch(() => setAuthenticationTokens(undefined))` discards the error. Refresh-token expiry is normal so I'm not asking for a notification, but please at least `console.error(error)` here so transient failures (network blip, Keycloak unreachable) leave a trace. Same pattern in the interval at line 335-339.

client/src/providers/AuthenticationContext/AuthenticationProvider.tsx
```diff
@@ -5,135 +5,408 @@ import { AuthenticationContext } from './context'
...
-        ? jwtDecode<IDecodedRefreshToken>(refreshToken)
-        : undefined
+      try {
+        decodedRefreshToken = refreshToken
+          ? jwtDecode<IDecodedRefreshToken>(refreshToken)
+          : undefined
+      } catch {
+        keycloak.clearToken()
+        setAuthenticationTokens(undefined)
+        return
```
krusche, line 344: Empty catch — please add `console.error(error)` here too. A `jwtDecode` throw on a refresh token is unusual enough to be worth a log line, even if the recovery (clear + log out) is correct.

client/src/providers/AuthenticationContext/AuthenticationProvider.tsx
```diff
@@ -219,12 +491,226 @@ const AuthenticationProvider = (props: PropsWithChildren) => {
...
+              credentialId: toBase64Url(passkeyCredential.rawId),
+              clientDataJSON: toBase64Url(passkeyCredential.response.clientDataJSON),
+              attestationObject: toBase64Url(passkeyCredential.response.attestationObject),
+              challenge,
+            }),
+          })
+
+          if (!response.ok) {
+            throw new Error(await getPasskeyErrorMessage(response, undefined, 'register'))
+          }
```
krusche, line 662: If `navigator.credentials.create()` (line 606) succeeds but this `save` POST fails, the OS authenticator has already stored the credential locally — but the user is told registration failed. They'll likely retry and end up with multiple orphan credentials in their password manager / iCloud Keychain.

Worth distinguishing the error message for this specific path, e.g. "Your device saved the passkey but the server rejected it. Please contact support to clean it up."

client/src/providers/AuthenticationContext/AuthenticationProvider.tsx
```diff
@@ -245,9 +731,17 @@ const AuthenticationProvider = (props: PropsWithChildren) => {
         })
       },
       researchGroups: researchGroups,
+      isPasskeySupported,
     }
     // eslint-disable-next-line @eslint-react/exhaustive-deps -- researchGroups/setAuthenticationTokens/readySignal/access_token are captured by reference inside callbacks that read the latest value at call time; recomputing the entire context on each token refresh would re-render every consumer
```
krusche: The rationale lists `researchGroups`/`readySignal`/`access_token` as "captured by reference," but they're actually in the dep array on lines 737-744. The only dep being suppressed is `setAuthenticationTokens` (a stable React setter, which is why the disable is fine). Please rewrite to reflect what's actually being suppressed and why.

client/src/components/Header/Header.tsx
```diff
@@ -5,28 +5,66 @@ import {
...
+  const navigate = useNavigate()
+  const isLoginModalOpen =
+    context.isReady && openLoginModal && !context.isAuthenticated && context.isPasskeySupported
+
+  const onLoginModalClose = () => {
+    if (isLoginModalOpen) void navigate('/', { replace: true })
+  }
+
+  const onPasswordLogin = () => {
+    void context.login('/dashboard')
```
krusche: Hardcoding `/dashboard` drops the deep-link target. The fallback-login effect in `AuthenticatedArea.tsx:191` correctly passes `${location.pathname}${location.search}${location.hash}` when passkeys are unsupported — please mirror that here so the user lands back where they started.

```ts
const onPasswordLogin = () => {
  void context.login(`${location.pathname}${location.search}${location.hash}`)
}
```
Same issue on line 163 (the unauthenticated header's standalone Login button).

client/src/pages/SettingsPage/components/PasskeySettings/PasskeySettings.tsx
```diff
@@ -0,0 +1,126 @@
...
+          No passkeys registered yet.
+        </Text>
+      )}
+      {!isLoadingCredentials && credentials.length > 0 && (
+        <Stack gap='sm'>
+          {credentials.map((credential) => (
+            <Paper withBorder p='sm' key={credential.id}>
+              <Group justify='space-between' align='center'>
+                <Stack gap={0}>
+                  <Text fw={500}>{credential.userLabel?.trim() ?? 'Unnamed passkey'}</Text>
```
krusche: `??` only falls back on null/undefined. If `userLabel` is whitespace-only, `.trim()` returns `''` (falsy but not nullish), so this renders as an empty `<Text>`. Use `||`:
```ts
{credential.userLabel?.trim() || 'Unnamed passkey'}
```

krusche (changes requested):
Deep review focused on the new passkey login flow. Most of the implementation is well-structured (per-user storage keys, residentKey-required registration, parseAccountCredentialsResponse handling both response shapes, CDP-based e2e). The findings below cluster around **error handling on the unhappy paths** — they are user-visible UX regressions or silent failure modes, not architectural problems.

## Must address

1. **Cancelled passkey prompt shows a scary generic error.** When the user dismisses the OS biometric dialog, `navigator.credentials.get/create` rejects with a `DOMException` (`NotAllowedError` / `AbortError`). `getPasskeyErrorMessage` doesn't inspect `error.name`, so the user sees "Something went wrong. Please try again later." Branch on the DOMException name and either suppress the toast on cancellation or show a neutral "Sign-in cancelled."

2. **Silent SSO probe failure leaves the user stuck.** If `keycloak.init` rejects in `initializeKeycloakSession` (Keycloak unreachable, CSP iframe block, misconfigured realm), the `catch` only `console.error`s and the `finally` still triggers `triggerReadySignal()`. UI flips to `isReady && !isAuthenticated` with no diagnostic. Please `showSimpleError` at minimum.

3. **Deep-link target is lost when the login modal opens.** The modal's password "Login" button and the unauthenticated header's "Login" button both call `context.login('/dashboard')`, hardcoding the redirect. When passkeys are *not* supported, the fallback-login `useEffect` in `AuthenticatedArea.tsx:191` correctly preserves `location.pathname + search + hash`. The modal path should do the same.

4. **Whitespace-only passkey labels render blank.** `credential.userLabel?.trim() ?? 'Unnamed passkey'` uses `??`, which only falls back on null/undefined. `''` (whitespace-only after trim) is falsy but not nullish, so the fallback never fires. Use `||`.

## Should address

5. **Orphan credential when `save` POST fails after `navigator.credentials.create` succeeds.** The authenticator already stored the credential locally, but the user is told "registration failed." Worth distinguishing this in the error message ("saved on your device but the server rejected it — please contact support") so the user doesn't try the same authenticator three more times.

6. **Token-refresh failures are debuggable-but-silent.** A few `.catch` blocks (`:209-211`, `:335-339`) clear the session without logging the error. Refresh-token expiry is normal so I'm not asking for user-facing notifications, but please at least `console.error(error)` so transient/unexpected failures (network, Keycloak down) are debuggable.

7. **Stale eslint-disable rationale at `:736`.** The rationale claims `researchGroups`/`readySignal`/`access_token` are "captured by reference inside callbacks," but they are real entries in the dep array. The actually-suppressed dep is `setAuthenticationTokens`. Please rewrite the rationale to match.

## Test coverage gap

The e2e suite is solid for the golden path, but ~1100 lines of new TypeScript have no unit coverage. The cheapest, highest-leverage tests to add:
- `getPasskeyErrorMessage` — status × operation matrix, including JSON vs plain-text bodies and `DOMException` inputs. An off-by-one in `status >= 400 && status < 500` would ship silently today.
- `getDeviceName` — 8+ platform/browser branches; e2e only covers the CI runner.
- `shouldSkipPromptForStorageKeys` — date-rollover branch (next-day re-prompt) is untested; `getDate()` vs `getUTCDate()` regression would be invisible.

Also worth adding: an explicit e2e assertion that after registering, reloading does NOT re-show the prompt (currently proven only indirectly because the registration test also sets `DISABLE_PASSKEY_PROMPT_STORAGE_KEY`).

---

Nice work overall — these are mostly polish on a well-designed feature. Happy to discuss any of the above.

Claudia-Anthropica (changes requested):
@ITegs Took a fresh pass after the develop merge. The merge itself is clean — I checked the passkey files specifically and none of the safety work from earlier rounds (configurable rp.name, the AuthenticatedArea redirect race) got reverted, so no regressions from the merge. I'm holding off on re-approving though, because @krusche's deep review from today raises real unhappy-path issues I agree with: the scary generic error on a cancelled OS prompt, the lost deep-link when the login modal opens, the orphan-credential case when the save POST fails after navigator.credentials.create succeeds, and the whitespace-only label fallback. Those are worth landing before merge — I won't repeat them inline since he's already covered them well.

Claudia-Anthropica (changes requested):
@ITegs The new commits since my last pass are just the develop merge, SBOM regen and the 4.11.1 version bump — no functional code changed, and I confirmed the merge didn't revert any of the earlier safety work. CI is green again now that the SBOM hashes are fixed, nice. Still holding at request-changes though: the unhappy-path issues @krusche raised are all still open (cancelled-prompt error in passkey.ts, the dropped deep-link in Header.tsx, the orphan-credential case when the save POST fails, the empty-catch logging, and the whitespace label fallback). Once those threads are addressed I'm happy to re-approve.

Claudia-Anthropica (approved):
@ITegs All feedback I was tracking is addressed now, including the passkey unhappy-path fixes and the provider packaging cleanup. CI is green, approving.

PR #978, opened 2026-04-30
Fix #572: ignore arrow keys when editing text fields

### LINKED ISSUE #572: Review Applications: When pressing arrows while in a textfield navigation navigates away
@rabeatwork Noticed that when you write a comment and navigate in the textbox using keyboard arrows, you change the application. Skipping through applications using arrows needs to be disabled if a textbox is in focus.

krusche:
## Summary
- Skip arrow-key application navigation when focus is on an input, textarea, select or any contentEditable element
- Skip when modifier keys are held (so OS shortcuts still work)
- Switch from `window.onkeydown` to `addEventListener('keydown')` so we no longer clobber other handlers

Closes #572.

## Test plan
- [ ] Open Review Applications page
- [ ] Focus the comment textarea, press ←/→ — selection must not change
- [ ] Click outside the textarea, press ←/→ — selection cycles applicants as before
- [ ] Press Cmd/Ctrl+← in the textarea — caret moves, selection unchanged

## Bug Fixes
* Improved arrow key navigation in the applications sidebar to prevent interference with form controls (input fields, textareas, select elements) and editable content.
* Enhanced keyboard handling to ignore modifier key combinations during navigation.


Claudia-Anthropica (approved):
@krusche Clean fix. The early returns on form controls + contentEditable cover the reported case, the modifier-key bypass keeps OS word navigation working, and the switch from `window.onkeydown =` to `addEventListener` is a nice side improvement — the old assignment would silently clobber any other handler attached the same way. CI shows two infra failures (npm install in e2e, client Docker build) that aren't related to this diff.

Claudia-Anthropica (approved):
@krusche New commit just adds a regression test for #572. The fix itself is unchanged from my previous review and still looks good. The test pins the source via regex rather than exercising behavior, but that matches the existing `client/test/*.test.mjs` style in this repo. Approving.

Claudia-Anthropica (approved):
@krusche Both follow-up comments are addressed cleanly — `shiftKey` is now in the modifier guard, and the stale-closure concern is fixed by routing `onSelect`/`setPage`/state through refs that get re-synced each render, so the keydown listener no longer re-binds while still reading current values. Nice.

Claudia-Anthropica (approved):
@krusche Nice follow-up — the source-grep test got replaced with a real Vitest unit test that actually exercises `shouldIgnoreArrowKey` against modifier keys, input/textarea/select targets, and contentEditable. Filter logic extracted cleanly into its own module. Tests, lint, and typecheck all green locally.

PR #979, opened 2026-04-30
Fix #463: confirm before deleting feedback

### LINKED ISSUE #463: Add confirmation dialog and undo option for deleting feedback
Currently, when clicking the 🗑️ trash icon to delete a feedback entry for a proposal, the feedback is deleted immediately without any confirmation. This can lead to accidental deletions with no way to recover the feedback.

**Expected Behavior:**
- When the user clicks the trash icon, a confirmation dialog should appear (e.g., "Are you sure you want to delete this feedback?").
- Optionally, include an "Undo" feature after deletion (e.g., a snackbar notification with an "Undo" button) to allow users to quickly revert the deletion.

**Why This Matters:**
- Prevents accidental data loss
- Improves user experience and trust in the system

**Suggested Implementation:**
- Add a confirmation modal before deletion
- Optionally, implement a temporary soft delete with an undo option (e.g., 5-second delay before final deletion)

krusche:
## Summary
- Wrap the trash button in `ThesisFeedbackOverview` with the existing `ConfirmationButton` so deletion now requires a Cancel/Confirm step
- Reuses the established Mantine modal pattern already used elsewhere in the app (no new component)

Closes #463.

The optional undo/snackbar from the issue is intentionally deferred — the confirmation dialog already removes the accidental-click footgun.

## Test plan
- [ ] As a supervisor, open a thesis with feedback entries
- [ ] Click the trash icon — modal appears with "Cancel" and "Confirm"
- [ ] Cancel keeps the feedback row intact
- [ ] Confirm removes the row and shows the existing success toast


Claudia-Anthropica (approved):
@krusche Clean fix — the swap to `ConfirmationButton` matches the pattern already used in the other Thesis sections, props are wired correctly, and the unused `Button` import is gone. The build-client / e2e failures are unrelated (pre-existing `package-lock.json` desync on develop: `@tiptap/extension-bubble-menu`/`-floating-menu` 3.22.5 missing from the lock file), not caused by this PR. Approving.

.github/workflows/run_tests.yml
```diff
@@ -49,3 +49,27 @@ jobs:
...
+      - name: Set up Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: '24.15.0'
+          cache: 'npm'
+          cache-dependency-path: 'client/package-lock.json'
+
+      - name: Install client dependencies
+        working-directory: ./client
+        run: npm ci
```
Claudia-Anthropica, line 63: @krusche [high] Heads up — this new `npm ci` step fails on the same pre-existing tiptap lock desync I flagged last round (`@tiptap/extension-bubble-menu` and `@tiptap/extension-floating-menu` 3.22.5 missing from `client/package-lock.json`). It's broken on develop too, but adding this step here means run-tests/build now also goes red, in addition to build-client and e2e. Since you're already touching the lock for the vitest deps, easiest path is to regenerate it once (`npm install` in `client/`) so the missing optional-peer entries get backfilled — otherwise the new client test job can't pass on this PR.

In the `client/` directory of this repo, run `npm install` to regenerate `client/package-lock.json` so the missing `@tiptap/extension-bubble-menu@3.22.5` and `@tiptap/extension-floating-menu@3.22.5` entries (referenced as optional peer deps of `@tiptap/starter-kit`) get backfilled. Commit the updated `client/package-lock.json` so that `npm ci` succeeds in the new `run-tests / build` and existing `build-dev-container / build-client` jobs.

</details>

Claudia-Anthropica (changes requested):
@krusche Thanks for swapping the source-pin for a real RTL component test — that's a much better signal than a regex against the source. The Vitest+RTL scaffold (jsdom polyfills, providers wrapper, afterEach cleanup) all looks sensible. The blocker is that the new `npm ci` step in run-tests fails on the same pre-existing tiptap lock desync I flagged on the previous round, so the client test job can never go green here. Easiest fix is to run `npm install` once locally and commit the resulting lock — see inline.

.github/workflows/run_tests.yml
Claudia-Anthropica: @krusche Lock is back in sync, run-tests/build is green now including the new vitest job. Thanks.

Claudia-Anthropica (approved):
@krusche All previous feedback addressed — lock is back in sync, run-tests/build is green including the new vitest job. Approving.

Claudia-Anthropica (approved):
@krusche Re-reviewing after the develop merge — the net diff against develop is unchanged from the version I approved at 65b507b. New commits on the branch are just upstream catch-up (vitest scaffolding from #989, dep bumps, privacy text). Vitest run of the new test file passes locally and run-tests/build is green. Approving.

PR #980, opened 2026-04-30
Fix #729: avoid mobile scroll hijack on presentation overview

### LINKED ISSUE #729: Bugfix: Presentation Overview - Mobile Scroll does not work as intended
In the Mobile view, the presentation overscrolls under the header.

krusche:
## Summary
- Skip the custom `window.wheel` handler below the `md` breakpoint so mobile uses native page scroll
- Eliminates the bug where the calendar/list scrolled *under* the sticky page header on small screens

Closes #729.

## Test plan
- [ ] Open `/presentations` on a desktop browser, scroll with the mouse wheel — calendar follows wheel input as before
- [ ] Resize the browser to <768px (or use device emulation) — page scrolls natively, header stays in place, no overscroll under the header
- [ ] Switching between mobile and desktop viewports re-applies the correct behavior (handler is keyed off `isSmaller`)


Claudia-Anthropica (approved):
@krusche Clean, targeted fix — early-returning before registering the wheel listener and keying the effect on `isSmaller` is exactly the right shape, and the cleanup runs correctly when crossing the breakpoint either way. CI red is unrelated (`npm ci` fails on develop because @tiptap/extension-bubble-menu and -floating-menu are missing from the lockfile). Approving.

Claudia-Anthropica (approved):
@krusche Regression test added since my last approval — it's a source-string regex check rather than a runtime test, and `npm test` isn't wired into any CI workflow (only ./gradlew test runs), so it's effectively documentation-grade. The fix itself is unchanged and still good. CI red is the unrelated lockfile drift on develop. Approving.

Claudia-Anthropica (approved):
@krusche Lockfile sync looks right — adds the missing @tiptap/extension-bubble-menu and -floating-menu entries and hoists path-to-regexp, which is what `npm ci` was tripping on. CI is green now (deploy-dev-container still pending, unrelated). Re-approving.

Claudia-Anthropica (approved):
@krusche Test conversion looks right — moving it under `src/` so Vitest's `include` glob picks it up means it now actually runs in CI (`vitest run` step in run_tests.yml), which addresses my previous note that the legacy node:test file wasn't wired into the workflow. Verified the two assertions still pass locally. Fix itself is unchanged. Approving.

PR #981, opened 2026-04-30
Fix #764: make topic research group read-only for non-admins

### LINKED ISSUE #764: Bug: Cant change research group of own topic
When suggesting my own topic, I cant change the research group afterwards.

Steps to reproduce:

- suggest your own topic
- fill out everything select research group "A"
- edit suggested topic, select research group "B"
- Edit shows "successful, but topic is still not in the new group

krusche:
## Summary
- Issue #764 reported that editing a "suggested" topic appeared to accept a new research group but silently kept the old one. Root cause: every non-admin user belongs to exactly one research group server-side (`User.researchGroup` is a single FK; `CurrentUserProvider.assertCanAccessResearchGroup` only permits the caller's own group), so any non-admin pick other than the topic's existing group was rejected on save. The old UI hid this by clipping the dropdown to the current group, which made the field look broken.
- **Client (`ReplaceTopicModal`):** removed the "looks editable, then fails" affordance for non-admins. The research group is now a `Select` only for admins; everyone else sees a read-only `TextInput` pre-filled with the topic's current group (or the user's own group on the create flow) and a hint that only administrators can change it. The `/v2/research-groups` fetch is skipped entirely for non-admins. The auto-select-single-group behavior is now gated to the create flow so editing a topic no longer clobbers existing examiners.
- **Server (`TopicService.updateTopic`):** added a defense-in-depth ownership/access guard so that non-admin non-owners (i.e. not the creator, a listed supervisor, or a listed examiner) cannot edit topics outside their own research group via a direct API call. Added `Topic.getExaminers()` to support that check.
- **E2E:** `topic-workflow.spec.ts` updated to assert the read-only field + hint instead of the old disabled dropdown.

Closes #764. See [this comment on #764](https://github.com/ls1intum/thesis-management/issues/764#issuecomment-4459142202) for the user-facing explanation.

## Test plan
- [ ] As an **admin**, edit a topic — research group renders as a dropdown listing all non-archived groups; switching it persists after save.
- [ ] As an **admin**, create a new topic — research group dropdown is populated; if there is only one group in the system, it auto-selects and auto-fills its head as examiner (unchanged behavior).
- [ ] As a **non-admin who created the topic** (supervisor/advisor), edit it — research group renders as a read-only field showing the topic's current group with the "only administrators can change…" hint; saving other fields works, does not change the group, and does not clobber existing examiners.
- [x] As a **non-admin who is listed as supervisor/examiner** on someone else's topic, same read-only behavior; saving other fields works.
- [ ] As a **non-admin who is not listed on the topic**, attempting to PUT `/v2/topics/{id}` directly (curl/Postman) with a different `researchGroupId` returns 403 (server-side guard).
- [ ] As a **non-admin**, create a new topic — research group is read-only and pre-filled with the user's own research group; submit succeeds.


Claudia-Anthropica (approved):
@krusche Small, surgical fix — the ownership predicate matches the pattern used elsewhere (e.g. `TopicPage.checkIfUserIsExaminerOrSupervisor`) and the backend's `assertCanAccessResearchGroup` guards the actual mutation, so this is a safe UI unlock. The CI failures (`build-client` / `e2e`) are an unrelated lockfile mismatch (missing `@tiptap/extension-bubble-menu@3.22.5`) — not from this PR. LGTM.

Claudia-Anthropica (approved):
@krusche Test addition looks good — same pattern as the #759 regression test. Re-approving on the new SHA. (CI build-client / e2e failures are still the unrelated lockfile mismatch.)

Claudia-Anthropica (approved):
@krusche Nicely targeted follow-up — the early `if (topic && !currentUser) return` plus adding `canEditResearchGroup` and `currentUser?.userId` to the dep array is exactly what was needed to handle the auth-loading race. CI is green now too. LGTM.

Claudia-Anthropica (approved):
@krusche Nice follow-up — gating the auto-select-examiners branch on the create flow (`!topic`) is exactly right, and extracting the client rule into a unit-tested helper reads much better than the inline grep test. Server-side guard mirrors the client well. CodeRabbit already noted the examiner asymmetry between client/server (and the comment) — leaving that thread to drive that decision. LGTM.

Claudia-Anthropica (approved):
@krusche The new commit lands the missing examiner branch on the server side so the authz check now mirrors `canEditResearchGroup` on the client — exactly what coderabbit's thread asked for. Clean merge with develop, no other surprises. LGTM.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:
@krusche on it! Taking a look now :eyes:

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:
@krusche on it! Taking a look now :eyes:

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:
@krusche on it! Taking a look now :eyes:

PR #982, opened 2026-04-30
Fix #953: render GitHub usernames as profile links

### LINKED ISSUE #953: Direct link Github Profile
When users have to enter information about external services, e.g. Github profile, the platform should directly provide a link to direct towards the external service. (E.g. redirect to Github profile)

krusche:
## Summary
- Add `renderCustomDataValue` helper that maps `customData['GITHUB']` to `https://github.com/<user>` (or accepts a full URL)
- Use it in the two places customData is rendered: `ApplicationData` and `ThesisStudentInfoSection`
- Robust against `@username`, full URLs, and invalid input (falls back to plain text)

Closes #953.

## Test plan
- [ ] Open an application whose author entered a GitHub username — it now renders as a link to `https://github.com/<user>` (new tab)
- [ ] Same on the Thesis page student info
- [ ] Empty/invalid GitHub values still render as plain text (no broken link)


client/src/utils/customDataLink.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { ReactNode } from 'react'
+
+const linkBuilders: Record<string, (value: string) => string | null> = {
+  GITHUB: (value) => {
+    const trimmed = value.trim().replace(/^@/, '')
+    if (!trimmed) {
+      return null
+    }
+    if (/^https?:\/\//i.test(trimmed)) {
+      return trimmed
```
Claudia-Anthropica: @krusche [low] The full-URL branch passes through *any* `https?://` value, not just github.com. Since `customData[GITHUB]` is rendered to other users (supervisors viewing applications, etc.), a student could enter `https://something-else.example/...` and it would render as a clickable link with the "Github" label. The link text is the URL so it's not really covert phishing, but the helper named `GITHUB` probably shouldn't open redirect anywhere. Consider parsing the URL and only accepting `github.com` (and maybe extracting the username path).

In `client/src/utils/customDataLink.tsx`, the `GITHUB` linkBuilder's full-URL branch (`if (/^https?:\/\//i.test(trimmed)) { return trimmed }`) accepts any http(s) URL, not just GitHub URLs. Replace it with logic that uses `new URL(trimmed)` (in a try/catch), checks `host === 'github.com' || host === 'www.github.com'`, and returns the canonical `https://github.com/<pathname-without-leading-slash>` form; return null otherwise so it falls back to plain text.

</details>

Claudia-Anthropica (approved):
@krusche Nice small helper, does the job cleanly. One low-prio note inline about the full-URL branch — Copilot already covered the rel/regex/truncation/test points. The two CI failures (`build-client`, `e2e`) are a pre-existing lockfile mismatch on develop (`@tiptap/extension-bubble-menu` / `extension-floating-menu` missing from `package-lock.json`), unrelated to this PR.

Claudia-Anthropica (approved):
@krusche Test addition follows the established node:test pattern in client/test/ — covers the documented GITHUB derivation rules. Re-approving; the rel/regex/truncation/full-URL points from the prior round are still on the table but I won't repeat them. CI failures are the unrelated repo-wide npm ci issue.

Claudia-Anthropica (approved):
@krusche Latest commit addresses everything from the prior round — the full-URL branch is now host-locked to github.com, the username regex matches GitHub's actual rules, the link is wrapped in a truncating Text, and the rel attribute is correct. Nice tightening. Approving.

Claudia-Anthropica (approved):
@krusche Clean Vitest conversion — exporting `buildGithubUrl` to test the URL derivation directly is a much better signal than the old source-grep, and the case coverage (full URL gating, invalid usernames, edge inputs) lines up with the production rules. Tests pass, CI is green, prior review feedback is all addressed. Approving.

PR #984, opened 2026-04-30
Feature #523: search research groups by head's name

### LINKED ISSUE #523: Enable Search by Head User Name in ResearchGroupAdminPage
Current Behavior:
The search functionality in the ResearchGroupAdminPage does not support filtering by the name of the head user.

Expected Behavior:
Admins should be able to search research groups by the name of the head user in addition to the group name.

Rationale:
Improves usability, especially in environments with many research groups, and helps quickly locate the responsible head user.

krusche:
## Summary
- LEFT JOIN the head user in the native search query and match `first_name`, `last_name`, and the concatenated full name in addition to the group's own name/abbreviation
- Add an explicit `countQuery` so pagination remains correct with the join

The frontend already forwards the search term as-is, so no client change is needed.

Closes #523.

## Test plan
- [ ] On `/admin/research-groups`, search for a head's first or last name — matching groups appear
- [ ] Existing search by group name/abbreviation still works
- [ ] Pagination works for searches that match many groups (count query)


Claudia-Anthropica (approved):
@krusche Clean and focused change — LEFT JOIN is correct (head_user_id is unique so no duplicates), the explicit countQuery keeps pagination consistent, and the integration test covers the new path. The e2e CI failure is an unrelated `npm ci` lockfile mismatch on `@tiptap/extension-bubble-menu` / `@tiptap/extension-floating-menu` that exists on develop, not from this PR. Approving.

Claudia-Anthropica (approved):
@krusche The lockfile sync pulls in just the missing @tiptap/extension-bubble-menu / -floating-menu peer entries and CI is now fully green. Server change is unchanged from my last review — still LGTM. Approving.

Claudia-Anthropica (approved):
@krusche Nice follow-up — replacing the LEFT JOIN with a gated EXISTS keeps the listing path free of the users join when no search is provided, and the count query stays in sync. Still LGTM.

server/src/test/java/de/tum/cit/aet/thesis/controller/ResearchGroupControllerTest.java
```diff
@@ -110,6 +110,35 @@ void getResearchGroups_IncludeArchived_ShowsArchivedGroups() throws Exception {
...
+			// At least one result, and groupA must be among them.
+			assertThat(json.get("content").size()).isGreaterThanOrEqualTo(1);
+			boolean found = false;
+			for (JsonNode entry : json.get("content")) {
+				if (groupA.toString().equals(entry.get("id").asText())) {
+					found = true;
+					break;
+				}
+			}
+			assertThat(found).as("expected to find group %s when searching by head name", groupA).isTrue();
```
krusche: **[Medium]** Test does not prove the head-name filter is applied

The assertions only check that `groupA` appears and that there is at least one result. They never verify that `groupB` (whose head's universityId is different from the search term) is excluded. If the search parameter were silently dropped server-side, this test would still pass — `groupA` would simply be in the unfiltered result set with very high probability.

For the test to actually demonstrate the new functionality, capture `groupB` and assert that it is NOT returned:

```java
UUID groupB = createTestResearchGroup("Other Group", headB.universityId());
// ...
boolean foundA = false;
boolean foundB = false;
for (JsonNode entry : json.get("content")) {
    String id = entry.get("id").asText();
    if (groupA.toString().equals(id)) foundA = true;
    if (groupB.toString().equals(id)) foundB = true;
}
assertThat(foundA).as("groupA must match the head's name").isTrue();
assertThat(foundB).as("groupB must NOT match — its head has a different name").isFalse();
```

A secondary nit on the comment at lines 115–117: it claims `TestUser sets both first and last name to the universityId`. The actual mechanism is `generateTestAuthenticationHeader` putting the universityId into the JWT `given_name`/`family_name` claims, which `updateAuthenticatedUser` then syncs into the DB on first request. Worth clarifying so the next reader doesn't go hunting for a setter.

krusche (commented):
## Summary

The EXISTS rewrite is a clean, semantically correct refactor of the original `LEFT JOIN` and is symmetric across the data and count queries. PG 18.3 supports all syntax used (`CONCAT`, `ILIKE`, `EXISTS`, `ANY(CAST(:x AS UUID[]))`). Performance is fine: because `users.user_id` is the PK and `research_groups.head_user_id` is FK to it, the correlated subquery becomes a single PK index lookup per outer row that didn't already short-circuit on name/abbreviation — better than the previous `LEFT JOIN`.

### Issues found

**Medium**
- The new integration test does not actually prove the head-name match works — it only asserts that `groupA` appears, never that the non-matching `groupB` is excluded. The test would still pass if the search filter were silently dropped. See inline comment on line 139.
- Search-pattern wildcard escaping: a user typing `%` or `_` (or `\`) makes ILIKE match too broadly. Pre-existing for name/abbreviation, but this PR widens the surface to user names. See inline on line 20.

**Low / Nit**
- Query body is duplicated verbatim in `value` and `countQuery` — any future change must be made in two places. See inline on line 42.
- Performance note about absent `pg_trgm` index — informational only, no action required at current scale.

### Verified clean
- Count query is symmetric with the main query (both have the EXISTS clause).
- `searchQuery == null || searchQuery.isEmpty()` normalization in `ResearchGroupService.getAll`/`getAllLight` (lines 102–103 / 127–128) means the empty-string edge case does not hit the SQL — `IS NULL` short-circuits the WHERE clause. Whitespace-only strings are not normalized — minor pre-existing nit.
- Pagination: `totalElements` is driven by the explicit `countQuery`, which now correctly counts rows that match via head name. No off-by-one.
- No new SQL injection surface — `:searchQuery` is bound, not interpolated.
- No PG version concerns at 18.3.

Requesting changes mainly so the test gets a proper negative assertion; the rest are nits.

server/src/main/java/de/tum/cit/aet/thesis/repository/ResearchGroupRepository.java
```diff
@@ -16,12 +16,30 @@ public interface ResearchGroupRepository extends JpaRepository<ResearchGroup, UU
 @Query(value = """
 		SELECT r.* FROM research_groups r
 		WHERE (:searchQuery IS NULL OR r.name ILIKE CONCAT('%', :searchQuery, '%')
-			OR r.abbreviation ILIKE CONCAT('%', :searchQuery, '%'))
+			OR r.abbreviation ILIKE CONCAT('%', :searchQuery, '%')
+			OR EXISTS (SELECT 1 FROM users h WHERE h.user_id = r.head_user_id
```
krusche, line 24: **[Medium]** ILIKE wildcard characters in `searchQuery` are not escaped

`searchQuery` is bound as a parameter (so no SQL injection), but the ILIKE pattern metacharacters `%`, `_`, and `\` are passed through literally. A user typing `%` will match every group; `_` will match anything with a single character at that position. This was already true for `r.name` and `r.abbreviation`, but the PR doubles the attack surface by including three more text fields (first_name, last_name, full name).

Mostly a UX / mild-DoS issue rather than a security one (no information disclosure beyond what the search endpoint already permits), so it's fine to leave for a follow-up. Suggested fix in the service layer:

```java
String searchQueryFilter = (searchQuery == null || searchQuery.isBlank())
    ? null
    : searchQuery.replace("\\", "\\\\").replace("%", "\\%").replace("_", "\\_");
```

and then `ILIKE CONCAT('%', :searchQuery, '%') ESCAPE '\\'` in the query. Note: since you use `ILIKE` (case-insensitive), the existing `.toLowerCase()` in `ResearchGroupService` is redundant — drop it. Also `searchQuery.isEmpty()` should probably be `isBlank()` so a single space doesn't get sent as a wildcard match-all.

server/src/main/java/de/tum/cit/aet/thesis/repository/ResearchGroupRepository.java
```diff
@@ -16,12 +16,30 @@ public interface ResearchGroupRepository extends JpaRepository<ResearchGroup, UU
...
+				AND (h.first_name ILIKE CONCAT('%', :searchQuery, '%')
+					OR h.last_name ILIKE CONCAT('%', :searchQuery, '%')
+					OR CONCAT(h.first_name, ' ', h.last_name) ILIKE CONCAT('%', :searchQuery, '%'))))
+			AND (CAST(:heads AS UUID[]) IS NULL OR r.head_user_id = ANY(CAST(:heads AS UUID[])))
+			AND (CAST(:campuses AS TEXT[]) IS NULL OR r.campus IS NULL
+			OR r.campus = ANY(CAST(:campuses AS TEXT[])))
+			AND (:includeArchived = TRUE OR r.archived = FALSE)
+		""",
+	nativeQuery = true)
```
krusche, line 46: **[Low/Nit]** Query body is duplicated verbatim between `value` and `countQuery`

The WHERE clause is now ~7 lines (with the new EXISTS), identical in both blocks. Any future change (add a filter, swap EXISTS for a JOIN, fix a bug) must be made in two places, and a missed update will silently desynchronize `totalElements` from `content`.

Options:
1. Extract the WHERE clause into a `private static final String WHERE_CLAUSE = """ ... """;` constant on a helper class and concatenate into both queries.
2. Move to a `JdbcClient` / `NamedParameterJdbcTemplate`-based custom repository fragment where the SQL is built once.
3. Leave it — the cost of divergence is low because both blocks are right next to each other and a reviewer will spot a discrepancy.

Not blocking, but mention because this is now the second WHERE clause and a third filter parameter will push past the readability cliff.

server/src/main/java/de/tum/cit/aet/thesis/repository/ResearchGroupRepository.java
```diff
@@ -16,12 +16,30 @@ public interface ResearchGroupRepository extends JpaRepository<ResearchGroup, UU
 @Query(value = """
 		SELECT r.* FROM research_groups r
 		WHERE (:searchQuery IS NULL OR r.name ILIKE CONCAT('%', :searchQuery, '%')
-			OR r.abbreviation ILIKE CONCAT('%', :searchQuery, '%'))
+			OR r.abbreviation ILIKE CONCAT('%', :searchQuery, '%')
+			OR EXISTS (SELECT 1 FROM users h WHERE h.user_id = r.head_user_id
+				AND (h.first_name ILIKE CONCAT('%', :searchQuery, '%')
+					OR h.last_name ILIKE CONCAT('%', :searchQuery, '%')
+					OR CONCAT(h.first_name, ' ', h.last_name) ILIKE CONCAT('%', :searchQuery, '%'))))
```
krusche: **[Nit]** Performance note (no change requested)

The correlated EXISTS evaluates `users` once per outer row that did not short-circuit on `r.name`/`r.abbreviation`. Since `head_user_id` is FK to `users.user_id` (PK), this is a single index lookup per group — cheap. With no `pg_trgm` index on `users.first_name`/`users.last_name`, the per-row ILIKE on those fields is a full string comparison, but only against one user per group, not against all users. This is materially better than the previous `LEFT JOIN`, which produced one row per `(group, user)` combination if `users` ever fanned out.

If the user table grows past ~100k rows AND this query becomes hot, add a `pg_trgm` GIN index on `users.first_name`/`users.last_name`. Until then, no action needed.

PR #985, opened 2026-04-30
Feature #524: include memberCount in ResearchGroupDto

### LINKED ISSUE #524: Include Member Count in ResearchGroupDto
Current Behavior:
The member count displayed in the top-right corner of each ResearchGroupCard is fetched via a separate backend call.

Expected Behavior:
Include the memberCount as an optional field in the ResearchGroupDto, populated in the existing call used to retrieve the list of research groups.

Rationale:
Reduces the number of backend calls, improving performance and simplifying frontend logic.

krusche:
## Summary
- Add nullable `memberCount` field to `ResearchGroupDto`
- List endpoint computes all counts in one grouped query (`countMembersGroupedByResearchGroup`) to avoid N+1 against `/members`
- Single-group endpoint uses `countByResearchGroupId`
- `ResearchGroupCard` consumes `props.memberCount` directly; the per-card `/members` fetch on mount is removed

Closes #524.

## Test plan
- [ ] `/v2/research-groups` JSON now includes a `memberCount` per group
- [ ] Research Group admin page renders the count badge without issuing a `/members` request per card (network panel: only the list call)
- [ ] Counts match for groups with several members and zero for empty groups
- [ ] Single-group endpoint `/v2/research-groups/{id}` also returns `memberCount`


Claudia-Anthropica (approved):
@krusche Clean fix for the N+1 — one grouped query for the list endpoint, single count for the detail endpoint, and the card just reads `props.memberCount`. Nice that this also incidentally corrects the old behavior where the badge was showing `page.content.length` (capped at the 50-row page size) instead of the total. Test covers both endpoints. LGTM.

PR #986, opened 2026-04-30
Feature #594: banner and footer label for non-production environments

### LINKED ISSUE #594: Add banner to test/dev environment
It should be super clear to users and testers whether they use a production or test / dev system.
We should add a banner (e.g. similar to Artemis test servers) and ideally we add a second line to the footer so that we know immediately which version is deployed (this can avoid lots of confusion during testing)

Image

Image

krusche:
## Summary
- New `ENVIRONMENT` runtime variable parsed by `GLOBAL_CONFIG` (`production` | `staging` | `test` | `dev`)
- New `EnvironmentBanner` rendered inside `AppShell.Header` in both `PublicArea` and `AuthenticatedArea`
- Header height automatically extends by the banner height so the existing main-content layouts keep their full viewport
- Footer renders the version with the environment suffix (e.g. `v1.0.0 · dev`)
- Production / unset environment: behavior unchanged

Closes #594.

## Test plan
- [ ] Run with `ENVIRONMENT=dev` (e.g. via `RUNTIME_ENVIRONMENT_VARIABLES` in `index.html`): red banner above the header and `· dev` next to the version
- [ ] `ENVIRONMENT=test`: orange banner
- [ ] `ENVIRONMENT=staging`: blue banner
- [ ] No `ENVIRONMENT` (or `production`): no banner, footer shows just `v…` like before
- [ ] Check both unauthenticated (`/about`) and authenticated layouts


client/src/components/EnvironmentBanner/EnvironmentBanner.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const label = labelByEnvironment[environment]
+
+  return (
+    <Group
+      gap='xs'
+      justify='center'
+      align='center'
+      bg={`${color}.1`}
+      c={`${color}.9`}
+      h={ENVIRONMENT_BANNER_HEIGHT}
```
Claudia-Anthropica, line 46: @krusche [low] The banner forces a fixed `h={28}` regardless of content, but the message is ~85 chars. On mobile widths (e.g. ~375px) the `Text` will either overflow horizontally or wrap to a second line and get vertically clipped by the fixed height. Consider switching `h` to `mih` so the banner grows on narrow screens, or trimming to just `${label} environment` and dropping the disclaimer trailer. If you do let the banner grow, remember `HEADER_HEIGHT` becomes responsive too.

In `client/src/components/EnvironmentBanner/EnvironmentBanner.tsx`, the `Group` has a fixed `h={ENVIRONMENT_BANNER_HEIGHT}` which clips long content on mobile widths. Replace `h={ENVIRONMENT_BANNER_HEIGHT}` with `mih={ENVIRONMENT_BANNER_HEIGHT}` so the banner can grow when the message wraps, or shorten the rendered text to just `${label} environment` and remove the disclaimer trailer to keep one line on small screens.

</details>

Claudia-Anthropica (approved):
@krusche Clean implementation overall. Copilot already flagged the substantive points inline (footer suffix accidentally rendering on `production`, `Container h='100%'` overflowing the now-extended header in both layouts, `ENVIRONMENT` missing from `ALLOWED_ENVIRONMENT_VARIABLES` in `generate-runtime-env.js`) — those should land before merge. Just one small mobile consideration from me, otherwise lgtm. The CI failures (`build-client`, `e2e`) look unrelated to this PR — they're tiptap lock-file mismatches on develop.

Claudia-Anthropica (approved):
@krusche All four review concerns addressed cleanly — header sizing now uses an inner `Box h={baseHeaderHeight}` so the banner doesn't push header content out, `ENVIRONMENT` is in the runtime allowlist, footer suffix is gated on non-production, and the banner uses `mih` + wrap-safe sizing for mobile. LGTM.

Claudia-Anthropica (approved):
@krusche The vitest conversion is a clear improvement over the previous file-grepping node:test — `isEnvironmentBannerVisible` is now exercised directly with a mocked `GLOBAL_CONFIG` and covers all four environment values plus the undefined case. Verified locally that the test passes; CI is green. Approving.

PR #987, opened 2026-04-30
Fix #948: explicit toggle for suggested-topic application visibility

### LINKED ISSUE #948: Bug: Applications without supervisor not visible or actionable despite dashboard count
**Description**
A thesis application assigned to the correct research group but with no supervisor is counted in the dashboard ("you have 1 unreviewed applications") but does not appear in the application list and cannot be accepted.

**Steps to reproduce**
1. Create a new thesis application
2. Assign the correct research group
3. Leave the supervisor field empty
4. Observe dashboard shows "1 unreviewed application"
5. Navigate to the application list — application is not visible and cannot be accepted

**Expected behavior**
Applications without a supervisor should still appear in the list and be reviewable/acceptable.

krusche:
## Summary
- Replace the implicit derivation of \`includeSuggestedTopics\` (which silently turned to \`false\` whenever the user picked specific topics) with an explicit Switch in the application filter UI
- Default is ON; the toggle's value is sent verbatim to the backend
- Aligns the list visibility with what the dashboard count already includes, fixing the "1 unreviewed application but nothing in the list" report

Closes #948.

## Test plan
- [ ] Reproduce the original scenario (application with research group + no supervisor) — application is visible in the list and the dashboard count matches
- [ ] Toggle "Include suggested topics" off — suggested-topic applications disappear immediately; turn it back on, they reappear
- [ ] With a topics filter selected and the toggle ON, both the selected topics' applications and suggested-topic ones are listed
- [ ] With a topics filter selected and the toggle OFF, only the selected topics' applications are listed


Claudia-Anthropica (approved):
@krusche Clean fix — making `includeSuggestedTopics` an explicit toggle (defaulting to true and forwarded verbatim) is the right call to make the list visibility match what the dashboard counter already includes. Diff is small, focused on the root cause, and the regex-based regression guard nails down the legacy NO_TOPIC-derivation. Approving as-is. Heads up that the build-client / e2e CI failures look unrelated — develop has the same npm ci lockfile mismatch on the @tiptap deps.

Claudia-Anthropica (approved):
@krusche Both follow-ups land cleanly: the rephrased description now matches the server-side `a.topic IS NULL` semantics, and dropping the legacy NO_TOPIC option in the Topic select removes the contradictory-state trap Copilot flagged. Re-approving.

Claudia-Anthropica (approved):
@krusche The Vitest conversion of the regression test is a strict improvement over the old grep-the-source approach — it actually renders the Switch and asserts the behaviour (default checked, undefined treated as on, click flips to false). Consistent with the new ConfirmationButton.test.tsx pattern, and CI is green. Re-approving.

PR #988, opened 2026-04-30
Fix #759: auto-scroll presentation overview to today on load

### LINKED ISSUE #759: Presentation Page does not show todays presentation

krusche:
## Summary
- After presentations load, scroll the overview to today's date heading
- If today has no presentation, fall back to the next upcoming day, then the first available date

The data was already loaded — today's presentation was just visually buried after earlier-dated entries, so users perceived it as missing. The auto-scroll surfaces it on initial load.

Closes #759.

## Test plan
- [ ] Schedule a presentation for today, reload `/presentations` — page scrolls to today's heading on load
- [ ] If today has no presentation but there is one tomorrow — page scrolls to tomorrow
- [ ] When all presentations are in the past relative to the loaded data set (edge case, query still returns them when they're today) — page scrolls to the latest available date
- [ ] Switching between research groups still re-targets correctly

## Summary by CodeRabbit

* **New Features**
  * Presentations list now automatically scrolls to a relevant date heading when loaded, intelligently selecting today's date, the next upcoming date, or the most recent past date.

* **Tests**
  * Added comprehensive test coverage for date selection logic to ensure correct behavior across various scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ls1intum/thesis-management/pull/988)


client/test/issue-759-todays-presentation-scroll.test.mjs
```diff
@@ -0,0 +1,107 @@
+import { describe, test } from 'node:test'
```
Claudia-Anthropica: @krusche [low] This file is named like a regression test but no CI workflow runs it — `run_tests.yml` only invokes `./gradlew test`, and `npm test` (which would pick up `client/test/*.test.mjs`) isn't wired into any workflow. So a future refactor of the auto-scroll effect will silently break the assertions here without anyone noticing. Either wire `npm test` into the client CI (most useful), or move the regression coverage into the existing Playwright e2e suite where it can actually fail a build.

In `client/test/issue-759-todays-presentation-scroll.test.mjs`, the regression test is not executed by any GitHub Actions workflow (`run_tests.yml` runs only `./gradlew test`, and `npm test` is not invoked). Add a step to `.github/workflows/run_tests.yml` (or a new client-tests job) that runs `cd client && npm ci && npm test`, so this `node:test` file actually runs on PRs. Alternatively, replace this source-grepping test with a Playwright e2e that loads `/presentations` and asserts the page is scrolled to today's heading on initial load, since `e2e_tests.yml` already runs.

</details>

Claudia-Anthropica (changes requested):
@krusche Concur with the two Copilot findings on `PresentationOverviewPage.tsx` — the past-only fallback going to the earliest date contradicts the test plan, and the effect re-firing on every `onDelete`/`onUpdate` will yank users back to today every time they edit a card. Both should be addressed before merge. One small note on the new regression test inline. The build-client / e2e CI failures are a pre-existing `package-lock.json` mismatch on develop (`@tiptap/extension-bubble-menu` and `@tiptap/extension-floating-menu` missing from the lock file) — unrelated to this PR.

Claudia-Anthropica (approved):
@krusche Both Copilot findings addressed cleanly — past-only fallback now picks the most recent date, and the `lastScrolledGroupId` ref keeps the auto-scroll from yanking the viewport on `onDelete`/`onUpdate`. Approving.

Claudia-Anthropica (approved):
@krusche Refactor since my last approval is just extracting `pickTargetDate` into a pure helper with vitest coverage — clean, and it also retires the old `.mjs` regression test that wasn't wired into CI. The race condition CodeRabbit flagged on the group-switch path is real but pre-existed my prior approval, so deferring rather than re-raising — fix in a follow-up if you want. CI build failures are the same pre-existing develop lock-file mismatch on `@tiptap/extension-bubble-menu` / `@tiptap/extension-floating-menu`, unrelated to this PR. Re-approving.

Claudia-Anthropica (approved):
@krusche New commit just takes develop's package-lock.json verbatim — exactly the lock-file mismatch I flagged before, no product code touched. build-client / build-server are green now. Re-approving.

PR #1017, opened 2026-05-11
Upgrade ESLint to v10 and adopt @eslint-react

krusche:
## Summary
- Migrates the client from ESLint **9.39.4 → 10.3.0** ahead of the ESLint 9 EOL on 2026-08-06.
- Replaces **`eslint-plugin-react@7.37.5`** (no ESLint 10 release — [maintainer-confirmed incompatibility](https://github.com/jsx-eslint/eslint-plugin-react/issues/3977), upstream fix [PR #3979](https://github.com/jsx-eslint/eslint-plugin-react/pull/3979) still open) with **`@eslint-react/eslint-plugin@5.7.5`** — a modern, actively-maintained linter built for ESLint 10 with native React 19 support.
- Audits and strengthens the rule set: removes silenced rules that defeated type safety, adds type-aware bug-catching rules from `typescript-eslint`, and adds code-quality rules covering common JS/TS footguns.

## Toolchain changes
| Package | Before | After | Notes |
|---|---|---|---|
| `eslint` | 9.39.4 | **10.3.0** | Node ≥20.19.0; project is on ≥24.7.0 ✅ |
| `@eslint/js` | 9.39.4 | **10.0.1** | Match major |
| `eslint-plugin-react` | 7.37.5 | **removed** | Incompatible with ESLint 10 |
| `@eslint-react/eslint-plugin` | — | **5.7.5** added | React 19-native, peer `eslint ^10.3.0` |
| `@eslint/compat` | 2.1.0 | **removed** | FlatCompat shim no longer needed |
| `@eslint/eslintrc` | 3.3.5 | **removed** | Same — native flat config used throughout |
| `@typescript-eslint/*` | 8.59.2 | unchanged | Already supports ESLint 10 |

The ESLint `ajv@8` workaround patch was regenerated for 10.3.0; the eslintrc patch was deleted with its package.

## Rule set audit
**Removed silencing** of `@typescript-eslint/no-explicit-any` — disabling it defeated the whole point of using TypeScript.

**Added (type-aware — requires `parserOptions.project`):**
| Rule | Severity | Catches |
|---|---|---|
| `await-thenable` | `error` | Awaiting non-Promises (always wrong) |
| `no-floating-promises` | `warn` | Unhandled async — promote to `error` after backlog |
| `no-misused-promises` | `warn` | Async event handlers swallowing errors |
| `no-unnecessary-type-assertion` | `warn` | Dead `as Foo` casts |
| `prefer-nullish-coalescing` | `warn` | `\|\|` → `??` for null/undefined |
| `prefer-optional-chain` | `warn` | Modern syntax |

**Added (code quality):**
`consistent-type-imports` (auto-fixed 315 imports to `import type` for better treeshaking), `no-non-null-assertion`, `no-explicit-any` (warn), `eqeqeq: ['error', 'smart']`, `no-throw-literal`, `no-implicit-coercion` (with `boolean: false` so `!!x` keeps TS narrowing in `&&` chains), `no-console`.

**Intentionally off:**
- `no-undef`/`no-redeclare` — TypeScript already enforces both; `no-redeclare` rejects legitimate function overloads.
- `strict-boolean-expressions`, `return-await` — too noisy / stylistic.

## Latent bugs fixed by the stricter linter (12)
- `requests/handler.ts` — removed duplicate `else if (status === 404)` (unreachable dead branch).
- `DropDownMultiSelect.tsx` — `let filteredData = data` always overwritten → `const` ternary.
- `ThesisCommentsList.tsx` / `CalendarCarousel.tsx` — `<>` as map root needing a key → `<Fragment key={…}>`.
- `utils/format.ts` — `getInterviewStateColor` called `useMantineColorScheme()` from a non-hook function (rules-of-hooks violation). Refactored to accept `isDark: boolean`; 3 call sites updated.
- `TopicCardGrid.tsx` — hook was called conditionally (`gridContent ?? useTopicsContext()`). Now always calls `useContext` then chooses.
- `InterviewBookingPage.tsx` — hoisted PascalCase function `SlotInformation` to a real module-scope component.
- `CollapsibleDateCard.tsx`, `SlotItem.tsx`, `CollapsibleTopicElement.tsx` — removed JSX IIFEs flagged by React Compiler rules.
- `ResearchGroupAdminPage.tsx` — moved `key` before spread props (avoids JSX deopt).
- `AuthenticatedArea.tsx`, `ResearchGroupSettingPage.tsx` — `==`/`!=` → `===`/`!==`.
- `EmailTemplateEditPage.tsx` — removed `async`/`await` from two functions that call `doRequest` with a callback (the callback overload returns `() => void`, not a Promise).

## Remaining warnings
418 warnings remain (`0 errors`) — visible nudges that won't block commits. Top categories: `prefer-nullish-coalescing` (~120), `set-state-in-effect` (~95), `exhaustive-deps` (~81), `no-floating-promises` (~53), `no-misused-promises` (~33). These form a prioritised cleanup backlog and will be addressed in follow-up commits.

## Test plan
- [x] `npx eslint --version` reports `v10.3.0`
- [x] `npm run lint` exits `0 errors`
- [x] `npx tsc --noEmit` clean
- [x] `npm install` resolves with no peer-dep warnings
- [ ] Pre-commit hook: stage a `.tsx` change and verify `npx eslint --fix` still works
- [ ] `npm run dev` and click through one route — verify webpack pipeline picks up the new config


Claudia-Anthropica (approved):
@krusche Solid migration — the rule-set rationale is clear, and the real wins here are the latent bugs the stricter linter surfaced: the rules-of-hooks fix in format.ts, the conditional-hook fix in TopicCardGrid, the SlotInformation hoist, and the duplicate 404 branch in handler.ts. Verified locally: 0 errors, tests pass, tsc clean.

Claudia-Anthropica (approved):
@krusche The follow-up commits look good: the `||` → `??` autofix correctly carves out the `false` short-circuit cases with inline disables, the `use()` / direct `<Context>` migration is clean, and the three rule disables (set-state-in-effect, purity, naming-convention-ref-name) all have defensible rationale inline. Lint is 0 errors / 213 warnings, types are clean — happy to approve.

Claudia-Anthropica (approved):
@krusche Partial follow-up looks good — the `void` wraps for fire-and-forget async handlers are the right call for `no-misused-promises`, the documented `exhaustive-deps` disables explain the intent at each site, and the GanttChartTicks fix to capture `ticksRef.current` before the cleanup closure (plus `replaceChildren()` instead of `innerHTML = ''`) is a nice cleanup along the way. Approving.

Claudia-Anthropica (approved):
@krusche The final wrap-up commit looks clean — dropping `async` from non-awaiting `doRequest`-callback wrappers and wrapping async event handlers in `() => void fn()` is exactly the right call. Spot-checked the renamed-to-sync functions (fetchMyBooking, bookSlot, inviteInterviewees, the ResearchGroupMembers handlers, etc.) — none of their call sites await or `.then` the return, so the behavior is unchanged. Nice job closing out the floating/misused-promise backlog.

PR #1029, opened 2026-05-12
Development: Migrate client from npm to pnpm to speed up CI

krusche:
## Summary

- Switches `client/` from npm to **pnpm 11.1.1** (pinned via Corepack + `packageManager` field) to cut install time in CI and the Docker build. pnpm's content-addressable store + hard links make cold installs 2–3× faster and warm CI cache hits dramatically faster.
- **Side benefit:** pnpm's strict `node_modules` layout caught two missing direct dependency declarations that npm's flat hoisting was silently tolerating — `@tiptap/core` (used in `EmailTextEditor/Extension.ts`) and `linkifyjs` (used in `test/format.test.mjs`). Both are now declared explicitly.
- `patch-package` is retained as-is (works fine with pnpm via the existing `postinstall` hook).

## Key changes

- **`client/package.json`**: `packageManager: pnpm@11.1.1`, `engines.pnpm >=11`, added `preinstall: npx only-allow pnpm` guard, moved npm `overrides` into `pnpm.overrides` (with `parent>child` syntax for the nested `fork-ts-checker-webpack-plugin>schema-utils` entry).
- **`client/pnpm-workspace.yaml`** (new): `allowBuilds` entry for `@parcel/watcher`. Required so fresh CI installs don't fail on pnpm 11's `strictDepBuilds` default. (pnpm 11 reads settings from this file, not from `package.json`'s `pnpm` section — see [pnpm settings](https://pnpm.io/settings).)
- **`client/pnpm-lock.yaml`** (new) / **`client/package-lock.json`** (deleted): pnpm-format lockfile generated by `pnpm install`.
- **`client.Dockerfile`**: `corepack enable`, `HUSKY=0`, copies `patches/` before install so `patch-package` postinstall finds them, swaps `npm ci` → `pnpm install --frozen-lockfile` and `npm run build` → `pnpm build`.
- **`.github/workflows/run_tests.yml`** + **`.github/workflows/e2e_tests.yml`**: added `corepack enable` step, switched `actions/setup-node` `cache: 'pnpm'` with `pnpm-lock.yaml` path, replaced `npm ci` / `npx ...` / `npm run ...` with `pnpm install --frozen-lockfile` / `pnpm exec ...` / `pnpm dlx ...` / `pnpm ...`.
- **`execute-e2e-local.sh`**: pnpm equivalents throughout; added a corepack fallback check at the top.
- **`CLAUDE.md`**: client command examples updated to pnpm.

## Renovate / Dependabot

Both auto-detect `pnpm-lock.yaml`. The `package-ecosystem: "npm"` entry in `.github/dependabot.yml` is the correct ecosystem identifier for both npm and pnpm — no config change needed.

## Local verification

- [x] `pnpm install --frozen-lockfile` — clean install, no warnings (~3.5s)
- [x] `pnpm exec tsc --noEmit` — passes (after adding `@tiptap/core`)
- [x] `pnpm exec eslint src/` — passes
- [x] `pnpm exec vitest run` — 58/58 pass
- [x] `pnpm test:legacy` — 26/26 pass (after adding `linkifyjs`)
- [x] `pnpm build` — webpack production build succeeds in ~6s
- [x] `docker build -f client.Dockerfile -t client-pnpm-test .` — 271 MB image builds and serves HTTP 200

## Test plan

- [ ] CI `run_tests` job passes; compare install step duration vs. previous npm-based runs on `develop`
- [ ] CI `e2e_tests` job passes; compare total wall time vs. baseline
- [ ] CI `build_docker` job publishes multi-platform image successfully
- [ ] After merge, confirm Renovate / Dependabot open subsequent PRs against `pnpm-lock.yaml`
- [ ] Watch first production deploy for any runtime regressions in the nginx-served bundle

## Rollback

Revert the merge commit — `package-lock.json` is restored from git history and Docker image tags are unchanged.

## Summary by CodeRabbit

* **Chores**
  * Switched the project from npm to pnpm across CI, Docker builds, scripts, docs, and developer tooling; added a preinstall check to enforce pnpm and added Docker ignore rules.
* **Bug Fixes / Improvements**
  * Presentation duration input now enforces a 2‑minute minimum, uses local typing state, and auto-saves with debounce to reduce repeated network requests and improve UX.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ls1intum/thesis-management/pull/1029)


client/package.json
```diff
@@ -39,21 +40,21 @@
...
-    "flatted": "3.4.2",
-    "brace-expansion": "5.0.6",
-    "path-to-regexp": "0.1.13",
-    "yaml": "2.9.0",
-    "uuid": "14.0.0",
-    "picomatch@<4": "2.3.2",
-    "picomatch@>=4": "4.0.4",
-    "fork-ts-checker-webpack-plugin": {
-      "schema-utils": "4.3.3"
+  "pnpm": {
```
Claudia-Anthropica: @krusche [high] pnpm 11 doesn't read overrides from `package.json`'s `pnpm` section — it reads them from `pnpm-workspace.yaml` (which is what your own comment on that file already says about settings). The lockfile confirms it: `fork-ts-checker-webpack-plugin` now resolves to `schema-utils@3.3.0` instead of the overridden `4.3.3`, and you get duplicates like `ajv@6.15.0` + `8.20.0` and `minimatch@3.1.5` + `10.2.5` that the old npm override used to dedupe. I verified by setting an obviously-wrong override (`react: 18.0.0`) under `pnpm.overrides` here — pnpm regenerated the lockfile unchanged and react stayed at 19.2.6. Moving the same key into `pnpm-workspace.yaml`'s `overrides:` and re-running `pnpm install` does apply it (and writes a top-level `overrides:` block to the lockfile, which is currently missing). Move the entire `overrides` block over and regenerate `pnpm-lock.yaml`.

In `client/package.json`, the `pnpm.overrides` block is silently ignored by pnpm 11 (which reads settings from `pnpm-workspace.yaml`). Move the entire `overrides` map (keys: `ajv`, `minimatch`, `serialize-javascript`, `lodash`, `follow-redirects`, `flatted`, `brace-expansion`, `path-to-regexp`, `yaml`, `uuid`, `picomatch@<4`, `picomatch@>=4`, `fork-ts-checker-webpack-plugin>schema-utils`) into `client/pnpm-workspace.yaml` under a top-level `overrides:` key (alongside the existing `allowBuilds:`), then delete the `pnpm` object from `client/package.json` and run `pnpm install` in `client/` to regenerate `pnpm-lock.yaml`. Verify the regenerated lockfile contains a top-level `overrides:` block and that `fork-ts-checker-webpack-plugin@9.1.0` now lists `schema-utils: 4.3.3` (not `3.3.0`).

</details>

Claudia-Anthropica (changes requested):
@krusche Solid migration overall — clean and well-explained. One real issue though: the `pnpm.overrides` you moved into `package.json` are being silently ignored by pnpm 11, so several of the version pins you had under the old npm `overrides` are no longer applied (schema-utils, ajv, minimatch at least). See the inline comment for details and the fix. Once that's moved into `pnpm-workspace.yaml` and the lockfile is regenerated, this should be good to go.

client/package.json
Claudia-Anthropica: @krusche Still unfixed — `pnpm.overrides` is still in `package.json` (lines 43-59) and `pnpm-workspace.yaml` still has only `allowBuilds`. The lockfile confirms it: no top-level `overrides:` block, `fork-ts-checker-webpack-plugin@9.1.0` still pulls `schema-utils: 3.3.0` (lockfile line 6908), and the `ajv@6/8`, `minimatch@3/10`, `brace-expansion@1/5` duplicates are all still there. The docs commit only touched `docs/DEVELOPMENT.md` — the override move never happened.

Claudia-Anthropica (changes requested):
@krusche The docs update is nice, but the override migration from my previous review still isn't done — `pnpm.overrides` is still under `package.json` and the lockfile still shows the old duplicates / unoverridden `schema-utils@3.3.0`. Once that's moved into `pnpm-workspace.yaml` and the lockfile regenerated, this is good to go.

Claudia-Anthropica (changes requested):
@krusche New commit looks good — auto-installing Playwright chromium on fresh checkouts is a nice quality-of-life touch. The pnpm.overrides migration from the earlier review is still open though: the block is still in package.json, pnpm-workspace.yaml still has only allowBuilds, and the lockfile still pulls schema-utils@3.3.0 into fork-ts-checker plus the ajv/minimatch/brace-expansion duplicates. Once that's moved over and pnpm-lock.yaml regenerated, this is good to merge.

Claudia-Anthropica (changes requested):
@krusche The debounce fix on PresentationSettingsCard looks good — pattern matches ApplicationReviewForm and the guards on isFinite/min should keep intermediate parse events from racing. Only blocker still open is the `pnpm.overrides` migration from the earlier review: the block is still under `package.json` and the lockfile still shows `fork-ts-checker-webpack-plugin@9.1.0` pulling `schema-utils@3.3.0` instead of the overridden `4.3.3`. Once that's moved into `pnpm-workspace.yaml` and the lockfile regenerated, this is ready to ship.

krusche:
@Claudia-Anthropica Good catch — confirmed and fixed in b8e28412. Same pnpm 11 gotcha as `allowBuilds`: workspace-level settings (including `overrides`) are only read from `pnpm-workspace.yaml`, not from `package.json`'s `pnpm` section. The previous lockfile was carrying `schema-utils@3.3.0`, `ajv@6.15.0`, `minimatch@3.1.5`, `brace-expansion@1.1.12` alongside their overridden versions.

Moved the full `overrides` block into `pnpm-workspace.yaml` and regenerated the lockfile — every previously-pinned package now appears only at the override version. `pnpm exec tsc --noEmit`, `pnpm exec eslint src/`, and `pnpm build` still pass.

client/package.json
Claudia-Anthropica: @krusche Looks good now — `pnpm-workspace.yaml` has the full `overrides:` block, the lockfile has a top-level `overrides:` section, and `fork-ts-checker-webpack-plugin@9.1.0` now pulls `schema-utils: 4.3.3`. The `ajv@6.x`, `minimatch@3.x`, `brace-expansion@1.x` duplicates are also gone. Thanks!

Claudia-Anthropica (approved):
@krusche All feedback addressed — overrides are now in `pnpm-workspace.yaml`, the lockfile picked them up correctly, and the schema-utils/ajv/minimatch/brace-expansion dedup is back. Nice migration overall.

Claudia-Anthropica (approved):
@krusche Doc-only delta since my last approval — the Native vs Docker table now accurately reflects the static-serve E2E flow (matches `execute-e2e-local.sh` and `e2e_tests.yml`). Looks good.

.dockerignore
```diff
@@ -0,0 +1,34 @@
...
+client/test-results
+client/blob-report
+
+# IDE / OS
+.idea
+.vscode
+.DS_Store
+
+# VCS / CI metadata that the runtime image doesn't need
+.git
```
Claudia-Anthropica: @krusche [high] Excluding `.git` here will break the server image. `server.Dockerfile:10` does `COPY .git ../.git` to feed gradle-git-properties, and the build context for both images is the repo root (`build_docker.yml:77,151`), so this root `.dockerignore` applies. CI is green on *this* PR only because `build-server` is gated on `server/**` changes (build_docker.yml:30-33) — the breakage will land on whatever server-touching PR merges next, failing at the COPY step.

Easiest fix: drop the `.git` line. If you want to keep the client image lean, move client-only exclusions into a `client.Dockerfile.dockerignore` so the server build context is unaffected.

In `.dockerignore` line 33, `.git` is excluded from the Docker build context, but `server.Dockerfile:10` does `COPY .git ../.git` (required by the gradle-git-properties plugin) and both Dockerfiles use the repo root as build context (see `.github/workflows/build_docker.yml` lines 77 and 151). Remove the `.git` entry from this root `.dockerignore`. If you want client-image-specific exclusions, put them in a `client.Dockerfile.dockerignore` file instead.

</details>

Claudia-Anthropica (changes requested):
@krusche The new `.dockerignore` is a nice cleanup, but it excludes `.git` and that will break the server image build the next time someone touches `server/**` (CI is green here only because `build-server` is gated on server changes). The `execute-e2e-local.sh` exit-code fix is clean — inline detail on the dockerignore.

.dockerignore
Claudia-Anthropica: @krusche Nice — `.git` is gone from the exclude list and the inline NOTE will save the next person from re-adding it. Server build is safe again. Thanks!

Claudia-Anthropica (approved):
@krusche All feedback addressed — `.git` removed from `.dockerignore` (with a NOTE so it doesn't come back), and PresentationSettingsCard now handles the empty-input case cleanly and skips prop-sync during in-flight saves. Nice work.

client/scripts/check-pnpm.js
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env node
+// Runs from the `preinstall` script in package.json. Aborts the install
+// when the user invoked anything other than pnpm (npm, yarn, bun, ...).
+//
+// Uses only the node binary that the invoking package manager already
+// spawned us with — no npx, no network fetch, no unpinned dependency.
+
+const userAgent = process.env.npm_config_user_agent || ''
```
Claudia-Anthropica, line 17: @krusche [critical] This is what's failing the e2e job. pnpm 11.1.1 does **not** set `npm_config_user_agent` when invoked with `--frozen-lockfile` (which is what CI does), so `userAgent` is `''`, `packageManager` becomes `''`, and the guard aborts with `Detected package manager: unknown`. Reproduced locally: with `pnpm install` user_agent is set (`pnpm/11.1.1 ...`), but with `pnpm install --frozen-lockfile` it's `undefined`. The local verification in the commit message used the non-frozen form and missed this. `npm_execpath` *is* set in both cases (points at `pnpm.cjs`), so check that as a fallback.

```js
const userAgent = process.env.npm_config_user_agent || ''
const execPath = process.env.npm_execpath || ''
const isPnpm = userAgent.startsWith('pnpm/') || /[\\/]pnpm(\.cjs|\.js)?$/.test(execPath)

if (!isPnpm) {
  // ... existing error ...
}
```

In `client/scripts/check-pnpm.js`, the `npm_config_user_agent` check fails under `pnpm install --frozen-lockfile` (pnpm 11.1.1 doesn't set that env var in that mode), which breaks CI. Replace the single-source detection with a check that accepts either `npm_config_user_agent` starting with `pnpm/` OR `npm_execpath` matching a `pnpm` binary path (e.g. ends in `pnpm`, `pnpm.cjs`, or `pnpm.js`). Verify with `cd client && rm -rf node_modules && pnpm install --frozen-lockfile` — it must succeed.

</details>

client.Dockerfile
```diff
@@ -1,14 +1,18 @@
...
+RUN corepack enable
 
-# Copy dependency files first for layer caching
-COPY client/package.json client/package-lock.json ./
-RUN npm ci
+# Copy dependency files (plus patches so patch-package can run in postinstall)
+# first for layer caching
+COPY client/package.json client/pnpm-lock.yaml client/pnpm-workspace.yaml ./
+COPY client/patches/ ./patches/
+RUN pnpm install --frozen-lockfile
```
Claudia-Anthropica, line 13: @krusche [critical] This is what's failing the `build-client` job. The preinstall hook is now `node scripts/check-pnpm.js`, but only `package.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`, and `patches/` are copied before `pnpm install` runs — `scripts/` isn't, so the install crashes with `Cannot find module '/app/scripts/check-pnpm.js'`. Add a `COPY client/scripts/ ./scripts/` line right after the patches copy.

```dockerfile
COPY client/package.json client/pnpm-lock.yaml client/pnpm-workspace.yaml ./
COPY client/patches/ ./patches/
COPY client/scripts/ ./scripts/
RUN pnpm install --frozen-lockfile
```

In `client.Dockerfile` around line 11, `pnpm install --frozen-lockfile` runs before `client/scripts/` is copied into the image, so the new `preinstall: node scripts/check-pnpm.js` hook in `client/package.json` fails with MODULE_NOT_FOUND. Add `COPY client/scripts/ ./scripts/` immediately after the existing `COPY client/patches/ ./patches/` line so the script exists when `pnpm install` runs. Verify with `docker build -f client.Dockerfile -t client-pnpm-test .` from the repo root.

</details>

Claudia-Anthropica (changes requested):
@krusche The new preinstall guard regressed CI — both `build-dev-container / build-client` and `run-e2e-tests / e2e` fail on this SHA. Two independent bugs in the same commit, details inline. The earlier `npx only-allow pnpm` was admittedly ugly but at least worked; the rewrite needs a small tweak.

client/scripts/check-pnpm.js
Claudia-Anthropica: @krusche Nice — verified locally that `pnpm install --frozen-lockfile` on a fresh checkout sets `npm_execpath` to the Corepack pnpm.cjs path even when `npm_config_user_agent` is empty, so `isPnpm` flips true via the execpath signal. `npm install` still gets rejected. Good fix.

client.Dockerfile
Claudia-Anthropica: @krusche Scripts dir is now copied before `pnpm install` — fixed.

Claudia-Anthropica (approved):
@krusche Both regressions from the previous review are fixed. The Dockerfile now copies `client/scripts/` before `pnpm install`, and `check-pnpm.js` uses `npm_execpath` as a secondary signal so pnpm under `--frozen-lockfile` (where `npm_config_user_agent` is empty in the Corepack flow) is still detected — verified locally with both `pnpm install --frozen-lockfile` (passes) and `npm install` (still rejected). Nice migration overall.

PR #1034, opened 2026-05-13
"Feat/add passkey login" branch attempt to pass the test

Lyynnnx:
In this Pull Request I tried to fix some issues after merge with "develop" branch in order to make all UI tests pass


client/src/declaration.d.ts
```diff
@@ -1,2 +0,0 @@
-declare module '*.png'
```
Claudia-Anthropica, line 1: @Lyynnnx [high] Deleting this file breaks the TypeScript build. `client/src/pages/AboutPage/AboutPage.tsx:2` does `import flowchart from './flowchart.svg'`, and without the `declare module '*.svg'` here, `npx tsc --noEmit` now fails with `TS2307: Cannot find module './flowchart.svg' or its corresponding type declarations.` Same risk for `*.png`. Please restore the file (or move the declarations into a kept `*.d.ts`).

Restore `client/src/declaration.d.ts` with the two original module declarations (`declare module '*.png'` and `declare module '*.svg'`). Verify by running `cd client && npx tsc --noEmit` and confirming there is no `TS2307` error on `src/pages/AboutPage/AboutPage.tsx`.

</details>

client/src/app/layout/AuthenticatedArea/AuthenticatedArea.tsx
```diff
@@ -221,10 +221,25 @@ const AuthenticatedArea = (props: PropsWithChildren<IAuthenticatedAreaProps>) =>
...
-                !item.groups || item.groups.some((role) => auth.user?.groups?.includes(role)),
-            )
+            // .filter(
+            //   (item) =>
+            //     !item.groups || item.groups.some((role) => auth.user?.groups?.includes(role)),
+            // )
+            .filter((item) => {
+              if (!item.groups) return true
+              const userGroups = auth.user?.groups?.map((g) => g.toLowerCase()) ?? []
+              const visibleLinks = links
```
Claudia-Anthropica: @Lyynnnx [high] `visibleLinks` is built (three chained `.filter` passes over the whole `links` array) and then never used — the callback returns `item.groups.some(...)` on the next line. This recomputes the full filtered link list once for every link on every render. Drop the `visibleLinks` block entirely; the outer pipeline already does the `display` and `hideFromGroups` filtering on lines 243-248.

In `client/src/app/layout/AuthenticatedArea/AuthenticatedArea.tsx` around lines 228-242, remove the unused `const visibleLinks = links.filter(...).filter(...).filter(...)` block inside the outer `.filter` callback. The callback should only compute `userGroups` and then `return item.groups.some((role) => userGroups.includes(role.toLowerCase()))`. Also delete the commented-out old filter on lines 224-227.

</details>

client/src/app/layout/AuthenticatedArea/AuthenticatedArea.tsx
```diff
@@ -327,11 +342,17 @@ const AuthenticatedArea = (props: PropsWithChildren<IAuthenticatedAreaProps>) =>
...
+                    requiredGroups.some((role) => 
+                      auth.user?.groups?.map(g => g.toLowerCase()).includes(role.toLowerCase())
+                    ) ? (
                       <ContentContainer size={size}>{children}</ContentContainer>
                     ) : (
-                      <Center className={classes.fullHeight}>
+                      <Center className={classes.fullHeight} data-testid="access-denied">
+                        <Stack align="center">
                         <h1>403 - Unauthorized</h1>
+                        <Text>Your roles: {user?.groups?.join(', ')}</Text>
```
Claudia-Anthropica: @Lyynnnx [low] Rendering the user's raw Keycloak group names on the 403 page is fine for debugging but probably shouldn't ship — it leaks role naming to anyone who hits an unauthorized route. If this was added to help diagnose test failures, gate it on a dev flag or drop it before merging.

In `client/src/app/layout/AuthenticatedArea/AuthenticatedArea.tsx` around line 354, either remove the `<Text>Your roles: {user?.groups?.join(', ')}</Text>` line or guard it behind a development-only check (e.g. `process.env.NODE_ENV !== 'production'`). The 403 page should not expose role names in production.

</details>

client/src/components/Header/Header.tsx
```diff
@@ -160,25 +162,27 @@ const Header = ({ opened, toggle, authenticatedArea, openLoginModal = false }: H
...
+                </Button>
+              )}
+              <Button onClick={() => void context.login('/dashboard')}>Login</Button>
+            </Group>
+          )
         )}
       </Flex>
       <Modal
-        opened={isLoginModalForcedOpen || isLoginModalOpen}
+        opened={isLoginModalForcedOpen ?? isLoginModalOpen}
```
Claudia-Anthropica: @Lyynnnx [medium] `??` is the wrong operator here. `isLoginModalForcedOpen` is a boolean (`openLoginModal && !context.isAuthenticated && context.isPasskeySupported`), so `false ?? isLoginModalOpen` evaluates to `false` and ignores `isLoginModalOpen` entirely. Today both flags move together so the modal still ends up correct, but the moment someone introduces another path that sets `isLoginModalOpen`, this silently breaks. Use `||`.

```suggestion
        opened={isLoginModalForcedOpen || isLoginModalOpen}
```

In `client/src/components/Header/Header.tsx` line 185, change `opened={isLoginModalForcedOpen ?? isLoginModalOpen}` back to `opened={isLoginModalForcedOpen || isLoginModalOpen}`. The left operand is a boolean expression, so `??` is semantically incorrect.

</details>

client/src/pages/InterviewBookingPage/InterviewBookingPage.tsx
```diff
@@ -92,11 +92,12 @@ const InterviewBookingPage = () => {
...
 
-      const interval = setInterval(() => {
-        auth.login()
-      }, 1000)
+      // const interval = setInterval(() => {
+      //   auth.login()
+      // }, 1000)
 
-      return () => clearInterval(interval)
+      // return () => clearInterval(interval)
```
Claudia-Anthropica: @Lyynnnx [medium] If the interval is intentionally gone, just delete it — don't leave a commented block plus a bare `return`. Same pattern is repeated in `CloseTopicButton.tsx`, `AuthenticatedArea.tsx`, `ThesisConfigSection.tsx`, `helpers.ts` and `playwright.config.ts` in this PR. The diff is hard to review with so much commented-out code sitting next to the new code.

In `client/src/pages/InterviewBookingPage/InterviewBookingPage.tsx`, remove the commented-out `setInterval` / `clearInterval` block and the bare `return` on line 100 — the useEffect should just call `auth.login()` once. Then sweep over this PR and delete the rest of the commented-out old code in `client/src/pages/ManageTopicsPage/components/CloseTopicButton/CloseTopicButton.tsx`, `client/src/app/layout/AuthenticatedArea/AuthenticatedArea.tsx`, `client/src/pages/ThesisPage/components/ThesisConfigSection/ThesisConfigSection.tsx`, `client/e2e/helpers.ts`, and `client/playwright.config.ts`.

</details>

client/src/pages/ThesisPage/components/ThesisConfigSection/ThesisConfigSection.tsx
```diff
@@ -74,13 +74,16 @@ const thesisDatesValidator = (
 
 const ThesisConfigSection = () => {
   const { thesis, access } = useLoadedThesisContext()
-  const hasAdminAccess = useHasGroupAccess('admin')
+  const { user } = useAuthenticationContext()
+  // const hasAdminAccess = useHasGroupAccess('admin') 
+  const hasAdminAccess = user?.groups?.map(g => g.toLowerCase()).includes('admin') ?? false
```
Claudia-Anthropica: @Lyynnnx [medium] Inlining `user?.groups?.map(g => g.toLowerCase()).includes('admin')` here defeats the whole point of having `useHasGroupAccess`. Every other call site in the codebase still uses the case-sensitive hook, so you've now got two different behaviours for what should be the same check. If the root cause is Keycloak returning mixed-case groups, fix it once in `client/src/hooks/authentication.ts:60` and keep `useHasGroupAccess('admin')` here.

In `client/src/hooks/authentication.ts` line 60, make `useHasGroupAccess` case-insensitive: change `user?.groups?.some((group) => groups.includes(group)) ?? false` to compare both sides lowercased, e.g. `user?.groups?.some((group) => groups.map((g) => g.toLowerCase()).includes(group.toLowerCase())) ?? false`. Then revert `client/src/pages/ThesisPage/components/ThesisConfigSection/ThesisConfigSection.tsx` line 79 back to `const hasAdminAccess = useHasGroupAccess('admin')` and drop the inline `useAuthenticationContext` import added on line 37. Apply the same simplification (lowercase comparisons inline) consistently in `client/src/app/layout/AuthenticatedArea/AuthenticatedArea.tsx`.

</details>

client/src/pages/ThesisPage/components/ThesisConfigSection/ThesisConfigSection.tsx
```diff
@@ -442,6 +445,7 @@ const ThesisConfigSection = () => {
                         color='red'
                         loading={anonymizeLoading}
                         onClick={() => void onDeleteThesisClick()}
+                        name="Anonymize Thesis"
```
Claudia-Anthropica: @Lyynnnx [low] `name="Anonymize Thesis"` on a Mantine `Button` isn't really useful — `name` only matters for form submit buttons, and the button text already provides an accessible name. If this is for Playwright targeting, use `data-testid` instead to stay consistent with `data-testid="access-denied"` you added in `AuthenticatedArea.tsx`.

In `client/src/pages/ThesisPage/components/ThesisConfigSection/ThesisConfigSection.tsx` line 448, replace `name="Anonymize Thesis"` with `data-testid="anonymize-thesis"` (or remove it if the Playwright test selects the button by its visible text). Use `data-testid` consistently across the codebase rather than `name`.

</details>

client/.husky/_/h
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env sh
```
Claudia-Anthropica: @Lyynnnx [medium] These four files (`client/.husky/_/h`, `pre-commit`, `pre-merge-commit`, `prepare-commit-msg`) are husky's auto-generated internal helpers — `client/.husky/_/.gitignore` is literally `*`, telling git to ignore everything in this directory. They get regenerated on every `husky install`, so committing them just creates merge conflict noise. Untrack them.

From the repo root run `git rm --cached client/.husky/_/h client/.husky/_/pre-commit client/.husky/_/pre-merge-commit client/.husky/_/prepare-commit-msg` and commit the deletions. These files are auto-generated by husky and are already excluded by `client/.husky/_/.gitignore` (contents: `*`), so they must not be tracked.

</details>

client/e2e/helpers.ts
```diff
@@ -6,12 +6,23 @@ import { Locator, Page, expect } from '@playwright/test'
...
+    `
+  }).catch(() => {});
+  await hideWebpackOverlay(page).catch(() => {});
   await page
     .locator('.mantine-Loader-root')
     .waitFor({ state: 'hidden', timeout: 30_000 })
     .catch(() => {
       // Loader may never appear if the page loads instantly
     })
+  await page.waitForTimeout(500);
```
Claudia-Anthropica, line 25: @Lyynnnx [medium] `waitForTimeout(500)` here adds a flat 500ms to every single `navigateTo` call across the whole e2e suite. That's an anti-pattern — Playwright's own docs flag it. If there's a specific signal you're waiting for, wait for that (a network response, an element becoming visible, a localStorage value being set). The `user-info` `waitForResponse` you added in `auth.setup.ts` is the right pattern; do the same here instead of a blind sleep.

In `client/e2e/helpers.ts` `navigateTo`, remove the `await page.waitForTimeout(500)` line. Identify what condition the timeout was masking (e.g. a specific network request, an element rendering, an animation finishing) and replace it with the corresponding explicit Playwright wait (`page.waitForResponse`, `locator.waitFor({ state: 'visible' })`, etc.). The existing `.mantine-Loader-root` wait above may already cover the case.

</details>

client/webpack.config.cjs
```diff
@@ -34,6 +34,7 @@ const config = (env) => {
       port: 3100,
       client: {
         progress: false,
+        overlay:false
```
Claudia-Anthropica: @Lyynnnx [nit] `overlay:false` — missing space after the colon, inconsistent with the surrounding object literal style. Also worth a short comment on why the overlay is being disabled (presumably so Playwright doesn't see it).

```suggestion
        overlay: false,
```

In `client/webpack.config.cjs` line 37, change `overlay:false` to `overlay: false,` to match the formatting of the surrounding object. Add a short comment explaining the overlay is disabled so it doesn't interfere with Playwright e2e tests.

</details>

Claudia-Anthropica (changes requested):
@Lyynnnx A lot of this PR is half-finished cleanup — old code commented out right next to the new code in several files, a dead `visibleLinks` block in `AuthenticatedArea`, and `client/.husky/_/*` files committed even though that directory has a `.gitignore` of `*`. The deletion of `client/src/declaration.d.ts` is the most concerning piece: `AboutPage.tsx` does `import flowchart from './flowchart.svg'`, and `npx tsc --noEmit` now fails with TS2307 on that line. Also `opened={isLoginModalForcedOpen ?? isLoginModalOpen}` in `Header.tsx` swaps `||` for `??` on a boolean — semantically wrong, even if the two flags happen to move together today. Details inline.

ITegs:
Fixed in 08284bafc8a8ad61d34cfc6a813d57f8ae19c8e1

PR #1047, opened 2026-05-15
Fix #754: persist application comments on fast applicant switch

### LINKED ISSUE #754: Comments on applications are only saved when focus is lost on the comment field
When writing a comment for an applicant in the comment field, there is a successful saved message below the comment text field.
However, when I then navigate to another applicant and back the comment is not saved correctly.

Currently, the comment is only saved correctly when I click somewhere outside the comments box so it loses focus. Then the correct comment is saved

krusche:
## Summary

Fixes [#754](https://github.com/ls1intum/thesis-management/issues/754) — comments typed into an applicant's review form were silently lost when the reviewer clicked another applicant in the sidebar before the auto-save debounce fired.

### Root cause

`ApplicationReviewForm` debounced the comment auto-save by 1 s. When the user switched applicants inside that window:

1. `useLayoutEffect` re-ran (the applicationId changed).
2. It called `form.setInitialValues(...)` then `form.reset()`, replacing `form.values.comment` with the new applicant's value.
3. The debouncer (which tracks `form.values.comment`) cancelled its pending timer — the typed text never made it to the `PUT`.

The workaround in the issue (click outside the textbox first) worked only because waiting after blur gave the debounce time to fire.

### Fix

`ApplicationReviewForm.tsx`:

- Track the latest typed comment and the most recent saved baseline in refs, kept in sync with `useEffect`.
- In the `useLayoutEffect` that handles application switching, add a cleanup that flushes any pending comment edit for the application being left, with two short-circuits to avoid redundant PUTs:
  - skip if pending text matches the closure baseline (no edits)
  - skip if it matches the latest saved value (debounced save already persisted it)
- The cleanup deliberately does **not** call `onUpdate`, so a late response can't yank the user back to the previous applicant.
- Tighten the debounce from 1000 ms → 500 ms so "Saved!" feedback appears sooner during normal typing — safe because the cleanup now protects against the previous race.

### Tests

- **Vitest** (`ApplicationReviewForm.test.tsx`): three regression specs covering the fast-switch flush, the no-op short-circuit, and the unchanged debounced-save path.
- **Playwright** (`application-comment-autosave.spec.ts`): logs in as supervisor, types a unique comment, immediately clicks another applicant inside the 500 ms window, hard-reloads the original applicant, and asserts the textarea reflects the server-persisted value.
- **Seed data** (`seed_dev_test_data.sql` §42): two dedicated NOT_ASSESSED applications (`c000-…-0040` Student2 and `c000-…-0041` Student3) on the supervisor's topic 1. Students 2 and 3 don't otherwise have NOT_ASSESSED apps in the seed, so the sidebar regex matches one row each, deterministically. Doesn't collide with apps 4/5 (`application-review-workflow`) or 12/13 (`interview-process-workflow`).

## Test plan

- [x] `cd client && pnpm test` — 67 vitest + 26 node tests pass
- [x] `cd client && pnpm exec tsc --noEmit` — clean
- [x] `cd client && pnpm lint` — clean
- [x] `cd server && ./gradlew test` — 805/805 pass
- [x] `cd server && ./gradlew spotlessCheck` — clean
- [x] `./execute-e2e-local.sh` — full Playwright suite, 226/226 pass (including the new spec)
- [x] Manual: reproduce in the browser — type a comment on applicant A, immediately click applicant B, return to A; comment is persisted (was lost before the fix)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed issue #754 where application comments were inadvertently lost when switching between applicants before the auto-save cycle completed.
  * Accelerated comment auto-save speed from 1000 milliseconds to 500 milliseconds for faster persistence and user feedback.
* **Tests**
  * Added comprehensive end-to-end and unit regression tests to verify comment auto-save behavior and integrity when switching between applicants.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ls1intum/thesis-management/pull/1047)

### DIFF
--- a/client/e2e/application-comment-autosave.spec.ts
+++ b/client/e2e/application-comment-autosave.spec.ts
@@ -1,0 +1,82 @@
+import { test, expect } from '@playwright/test'
+import { authStatePath, navigateToDetail } from './helpers'
+
+// Regression test for issue #754:
+//   Before the fix, typing a comment and immediately clicking on another
+//   applicant in the sidebar would silently drop the text — the 500ms
+//   auto-save debounce hadn't fired and the form-reset on the application
+//   switch wiped the unsaved value. The fix flushes any pending edit in
+//   the application-switch cleanup; this test guards that behaviour.
+//
+// The seeded applications c000-…-0040 (Student2) and c000-…-0041 (Student3)
+// are dedicated to this test (see seed_dev_test_data.sql §42) so we can
+// rely on Student2/Student3 being the only NOT_ASSESSED sidebar rows for
+// those students when the supervisor opens /applications.
+
+const APPLICATION_A_ID = '00000000-0000-4000-c000-000000000040' // Student2
+const APPLICATION_B_ID = '00000000-0000-4000-c000-000000000041' // Student3
+
+const COMMENT_TEXT = `issue-754 ${Date.now()} ${Math.random().toString(36).slice(2, 8)}`
+
+test.describe('Application comment auto-save — issue #754', () => {
+  test.use({ storageState: authStatePath('supervisor') })
+
+  test('persists a typed comment when the reviewer switches applicants before the debounce fires', async ({
+    page,
+  }) => {
+    // 1. Open applicant A's review page directly so we know exactly which
+    //    application's comment we're editing.
+    const studentAHeading = page.getByRole('heading', { name: /Student2 User/i })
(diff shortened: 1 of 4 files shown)


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:
@krusche roger that — review incoming!

PR #1075, opened 2026-05-22
Add admin dependency overview page and weekly vulnerability scan email

krusche:
## Summary

- Adds an admin-only `/admin/dependencies` page that lists every server (Maven) and client (npm) dependency from the bundled CycloneDX SBOMs, with per-component vulnerability lookups against [osv.dev](https://osv.dev/) and a 24h in-memory cache.
- Adds a Monday-morning cron (`thesis-management.dependency-scan.cron`, default `0 0 8 * * MON`) that emails the configured admin a vulnerability summary, with an upgrade recommendation when a newer GitHub release exists *and* high/critical CVEs are present.
- The page lets admins refresh vulnerability data on demand, trigger the email manually, and download either raw SBOM as JSON.
- Bundles the committed `server/sbom/bom.json` and `client/sbom/bom.json` into the jar via `processResources`, so the feature works without a separate build flag.

Mirrors the same feature in Artemis (`AdminSbomResource`, `SbomService`, `VulnerabilityService`, `ArtemisVersionService`, `VulnerabilityScanScheduleService`); the Angular UI is reimplemented in React + Mantine to match the existing project conventions.

## Implementation notes

- **Server package**: `de.tum.cit.aet.thesis.dependency` — controller + service + DTOs (incl. OSV API DTOs).
- **SBOM bundling**: copied via `processResources` into `sbom/server-sbom.json` and `sbom/client-sbom.json` on the classpath. Parsed once on first lookup and cached in-memory (parsing ~1.5 MB of JSON per request was a real perf cost).
- **OSV.dev**: batch API (1000 components per request) for IDs + per-ID detail fetches. CVSS v3/v4 vector heuristic for severity when no numeric score is published. Range matching uses `semver4j 3.1.0` (newly added dep).
- **Vulnerability cache**: in-memory `AtomicReference<CachedVulnerabilities>` with 24h TTL. Manual refresh evicts.
- **Version check**: GitHub `releases/latest` endpoint, repo configurable (`ls1intum/thesis-management` by default). 24h cache.
- **Email**: classpath Thymeleaf template at `templates/mail/vulnerability-scan-result.html`, rendered via the existing `MailConfig.templateEngine`. Bypasses the database-driven `email_template` table because this is a system-level notification, not per-research-group customizable. Respects `MailConfig.isEnabled()` — disabled mail dev path logs the rendered HTML instead of sending.
- **Auth**: every endpoint is `@PreAuthorize("hasRole('admin')")`; the nav entry is gated by the `admin` group.
- **Terminology**: no "frontend"/"backend" in code or messages — uses "server"/"client" per project memory.

## Config (`application.yml`)

```yaml
thesis-management:
  dependency-scan:
    cron: ${VULNERABILITY_SCAN_CRON:0 0 8 * * MON}
    admin-email: ${VULNERABILITY_SCAN_ADMIN_EMAIL:}
    github-repo: ${VULNERABILITY_SCAN_GITHUB_REPO:ls1intum/thesis-management}
    test-server: ${TEST_SERVER:false}
```

When `admin-email` is blank the cron logs a warning and skips. The scan is also skipped on `dev` profile and when `test-server=true`.

## Endpoints (all admin-only, prefix `/api/v2/admin/dependencies`)

| Method | Path | Description |
|---|---|---|
| GET | `/` | Combined server + client SBOM, 404 if neither bundled |
| GET | `/server` | Server SBOM only, 404 if not bundled |
| GET | `/client` | Client SBOM only, 404 if not bundled |
| GET | `/vulnerabilities` | Cached vulnerability data (24h TTL) |
| GET | `/vulnerabilities/refresh` | Bypasses cache, re-fetches from OSV |
| GET | `/version` | Current vs. latest GitHub release |
| POST | `/vulnerabilities/send-email` | Manual trigger; returns `{"sent": true\|false}` |

## Verification

Local checks:
- `./gradlew compileJava test` — 805 tests pass.
- `./gradlew processResources` — both SBOMs land in `build/resources/main/sbom/`.
- `pnpm exec tsc --noEmit` — 0 errors.
- `pnpm exec eslint src/` — 0 errors / 0 warnings.
- `pnpm build` — page emits as its own ~18 KiB chunk.

Pre-existing checkstyle warnings in `ResearchGroupService.java`/`ThesisService.java` are unrelated to this change; CI only runs `test` so they don't gate.

## Test plan

- [x] `docker compose up -d`; wait for Keycloak (`curl -sf http://localhost:8181/realms/thesis-management`).
- [x] `cd server && ./gradlew bootRun`; `cd client && pnpm dev`.
- [x] As an `admin` user, open `http://localhost:3100/admin/dependencies`.
- [x] Verify total/server/client counts populate and the table renders both Maven + npm components.
- [x] Filter by source (All / Server / Client) and free-text search; verify the count updates.
- [x] Click **Refresh vulnerabilities**; verify the page repopulates and the "last checked" timestamp updates.
- [x] Click **Send email** with `MAIL_ENABLED=false`; verify the success toast and the rendered HTML appears in the server logs at DEBUG. With `MAIL_ENABLED=true` and Mailpit (`http://localhost:8125`), verify the email lands with the summary table and the upgrade banner (if a newer release is published).
- [x] Click **Server SBOM** / **Client SBOM**; verify each downloads a valid CycloneDX JSON.
- [x] As a non-admin user, navigate to `/admin/dependencies`; verify the `AuthenticatedArea` blocks access.
- [x] If you want to exercise the cron path, temporarily set `VULNERABILITY_SCAN_CRON=*/2 * * * * *` and `SPRING_PROFILES_ACTIVE=prod`; revert before committing.

## Summary by CodeRabbit

* **New Features**
  * Admin-only "Dependencies" page and nav entry showing combined server+client SBOMs.
  * Interactive dashboard: search, source filter, "only vulnerable" toggle, severity-aware sorting, summary cards, and per-component vulnerability details.
  * Actions: refresh vulnerability data, send vulnerability-report email, and download server/client SBOM JSON files.
  * Alerts: missing SBOM warnings and update-available banner with release notes link.

* **Tests**
  * Added end-to-end and unit tests covering the new page and services.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ls1intum/thesis-management/pull/1075?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

server/src/main/java/de/tum/cit/aet/thesis/dependency/service/VulnerabilityService.java
```diff
@@ -0,0 +1,693 @@
...
+			String dbSeverity = vulnerability.databaseSpecific().severity().toUpperCase();
+			if (isValidSeverity(dbSeverity)) {
+				return normalizeSeverity(dbSeverity);
+			}
+		}
+
+		if (vulnerability.severity() != null && !vulnerability.severity().isEmpty()) {
+			for (OsvVulnerabilityDTO.OsvSeverityDTO severityEntry : vulnerability.severity()) {
+				if (severityEntry.type() != null && severityEntry.type().startsWith("CVSS")) {
+					Double cvssScore = parseCvssScore(severityEntry.score());
```
Claudia-Anthropica, line 559: @krusche [medium] OSV puts the CVSS *vector string* in `severity[].score` (e.g. `CVSS:3.1/AV:N/AC:L/...`), not a number — so `parseCvssScore()` here ends up doing `Double.parseDouble("CVSS:3.1/...")`, which always throws and returns null. Two consequences: `cvssScore` in the DTO is always null, and `mapCvssScoreToSeverity()` (the accurate CVSS v3.1 thresholds you wrote) never runs, so every severity falls back to the rough `estimateSeverityFromCvssVector()` heuristic. `extractCvssScoreFromOsvVulnerability()` has the same problem. If you want the precise scoring the code is clearly aiming for, parse the base score out of the vector (or use a CVSS library) before mapping.

In `server/src/main/java/de/tum/cit/aet/thesis/dependency/service/VulnerabilityService.java`, `parseCvssScore()` is called on OSV's `severity[].score`, but for CVSS_V2/V3/V4 entries OSV stores the CVSS *vector string* there, not a numeric score, so `Double.parseDouble` always fails and returns null (making `cvssScore` always null and `mapCvssScoreToSeverity` unreachable). Fix this by computing the numeric base score from the CVSS vector (e.g. via a CVSS scoring library) so both `determineSeverityFromOsvVulnerability` and `extractCvssScoreFromOsvVulnerability` can use it, falling back to the existing vector heuristic only when no vector is parseable.

</details>

client/src/pages/DependencyOverviewPage/DependencyOverviewPage.tsx
```diff
@@ -0,0 +1,608 @@
...
+  const [sortStatus, setSortStatus] = useState<DataTableSortStatus<IComponentRow>>({
+    columnAccessor: 'name',
+    direction: 'asc',
+  })
+
+  useEffect(() => {
+    let cancelled = false
+
+    void (async () => {
+      const [sbomRes, vulnRes, versionRes] = await Promise.all([
```
Claudia-Anthropica: @krusche [medium] `setLoadingSbom(false)` only runs after this `Promise.all` resolves all three requests, and `fetchVulnerabilities()` triggers the synchronous OSV scan on the server when the cache is cold (the per-ID detail fetches run sequentially with 30s `.block()` timeouts). So the first admin to open the page after a restart or the 24h expiry just sees a bare full-page `<Loader/>` until the whole scan finishes — the SBOM table is blocked even though its data already arrived. Render the table as soon as the SBOM resolves and let vulnerabilities populate independently with the `loadingVulnerabilities` indicator you already have; the table tolerates missing vuln data fine via `vulnerabilityIndex`.

In `client/src/pages/DependencyOverviewPage/DependencyOverviewPage.tsx`, the initial-load effect awaits `Promise.all([fetchCombinedSbom(), fetchVulnerabilities(), fetchVersionInfo()])` and only calls `setLoadingSbom(false)` afterwards, so the full-page loader stays until the (potentially slow, cold-cache) vulnerability scan completes. Decouple the requests: fetch the SBOM and set `loadingSbom` false independently of `fetchVulnerabilities()`, and gate the vulnerability data on the existing `loadingVulnerabilities` state so the component table renders immediately.

</details>

server/src/main/java/de/tum/cit/aet/thesis/dependency/service/ThesisManagementVersionService.java
```diff
@@ -0,0 +1,182 @@
...
+		cache.set(new CachedVersion(result, Instant.now()));
+		return result;
+	}
+
+	/**
+	 * Forces a refresh of version info from GitHub, bypassing the cache.
+	 *
+	 * @return the freshly fetched version info
+	 */
+	public ThesisManagementVersionDTO refreshVersionInfo() {
```
Claudia-Anthropica: @krusche [nit] `refreshVersionInfo()` isn't called anywhere — the controller's `/version` endpoint and the schedule service both use `getVersionInfo()`. Looks like dead code; drop it unless a refresh endpoint is coming.

In `server/src/main/java/de/tum/cit/aet/thesis/dependency/service/ThesisManagementVersionService.java`, the public method `refreshVersionInfo()` has no callers. Remove it.

</details>

Claudia-Anthropica (changes requested):
@krusche Nice, well-structured feature — clean package layout, and bundling the SBOMs via `processResources` plus the 24h caches are sensible choices. Two things worth a look beyond what Copilot and CodeRabbit already raised: the accurate CVSS numeric-score path is effectively dead because OSV puts a vector string in `score` (so severities only ever come from the rough heuristic), and the dependency page's full-page loader is held hostage by the cold-cache vulnerability scan. Details inline.

server/src/main/java/de/tum/cit/aet/thesis/dependency/service/VulnerabilityService.java
Claudia-Anthropica: @krusche Still open here — `parseCvssScore()` is unchanged, so `Double.parseDouble()` on OSV's CVSS *vector* string still always throws and returns null. That leaves `cvssScore` null in the DTO and the `mapCvssScoreToSeverity()` thresholds you wrote unreachable; severity always falls through to the `estimateSeverityFromCvssVector()` heuristic. Heads up: the new `mapsCvssScoreToSeverity` test only passes because it feeds `"score": "7.5"` for a `CVSS_V3` entry — real OSV puts the vector string in `score`, never a bare number, so that test doesn't exercise the production data shape. Parse the base score out of the vector (or use a CVSS lib) before mapping.

client/src/pages/DependencyOverviewPage/DependencyOverviewPage.tsx
Claudia-Anthropica: @krusche Nice — the decoupled per-resource fetches handle it. The SBOM table now renders as soon as the SBOM lands and vulns/version fill in independently, and the `cancelled` cleanup guard is a good touch. Resolving.

server/src/main/java/de/tum/cit/aet/thesis/dependency/service/ThesisManagementVersionService.java
Claudia-Anthropica: @krusche Fine to keep now that it's covered by tests and mirrors the vulnerability refresh path. Resolving.

Claudia-Anthropica (changes requested):
@krusche Thanks for the solid test coverage and the test-enabling refactors. Two of my three points from last round are handled — the decoupled fetches fix the cold-cache loader UX nicely, and the version-refresh method is fine to keep now that it's tested. Still open: the CVSS numeric-score path is dead because OSV stores a CVSS vector string in `severity[].score`, so `parseCvssScore()` always returns null — `cvssScore` is never populated and the precise threshold mapping never runs. The new `mapsCvssScoreToSeverity` test only passes because its fixture uses a bare `"7.5"`, which OSV never actually sends for a `CVSS_V3` entry. Left details in the thread.

krusche:
Thanks all for the careful reviews. Every actionable comment is addressed in `03fccab2`. Map of each open thread to the fix:

### Copilot

| # | Comment | Fix |
|---|---|---|
| 1 | Maven `3.1.5.RELEASE` / `4.2.12.Final` over-report (NPM `Semver` throws → catch returns `true`) | `VulnerabilityService.parseVersionLenient` — strips `.RELEASE`/`.Final`/`.GA`/`.SP*`/`.CR*` qualifiers and collapses 4-part Maven versions; on persistent failure returns `false` (conservative) instead of `true`. Test: `VulnerabilityServiceTest.mavenQualifierVersionDoesNotOverReportCves` |
| 2 | OSV batch catch logs only `e.getMessage()` | Switched to `log.error("Failed to query OSV API for batch starting at index {}", batchStartIndex, e)` so the stack trace is preserved |
| 3 | PR description says `/api/v2/admin/dependencies`, controller says `/v2/admin/dependencies` | Not a bug — `server.servlet.context-path: /api` is set in `application.yml`, so the controller-relative `/v2/...` lands at the full URL `/api/v2/...`. PR description is correct |
| 4 | `/vulnerabilities/refresh` should be POST (mutates cache) | Changed to `@PostMapping`; client request changed to POST. Test: `DependencyOverviewControllerTest.refreshRejectsGetMethod` (expects 405 on GET) |
| 5 | No controller tests | `DependencyOverviewControllerTest` (11 tests, added in `bd96b79e`) — covers admin vs non-admin gating, combined/server/client SBOM, vulnerabilities + refresh, version, send-email return shape |
| 6 | Mail sender catch makes scheduler falsely report `{sent: true}` | `VulnerabilityScanMailSender.send()` now returns `boolean`; `VulnerabilityScanScheduleService` checks it and returns `false` on failure |

### CodeRabbit

| # | Comment | Fix |
|---|---|---|
| 7 | Dual active state for `/admin` and `/admin/dependencies` | `AuthenticatedArea.tsx` now computes a `matchScore` per link and only highlights the longest-prefix match |
| 8 | Mail-send failures swallowed | Same as Copilot #6 |
| 9 | Admin email in INFO log (PII) | Removed `adminEmail` from the success log; message is now "Vulnerability scan completed. ... notification sent." |
| 10 | OSV multi-event ranges collapsed to last-seen pair | `VulnerabilityService.rangeContainsVersion` walks `events[]` sequentially and evaluates each closed `[introduced, fixed)` window independently. Tests: `multipleEventIntervalsAreEvaluatedSeparately` and `versionBetweenTwoVulnerableWindowsIsNotAffected` |
| 11 | UNKNOWN severity styled as LOW in email | Added explicit `LOW` and `unknown` branches to the `th:classappend` chain, plus a neutral-gray `unknown` CSS class |

### Claudia

| # | Comment | Fix |
|---|---|---|
| 12 | CVSS numeric-score path dead (OSV puts vector string in `score`) | New `CvssCalculator` implements the CVSS 3.x base-score formula per [CVSS v3.1 §7.1](https://www.first.org/cvss/v3.1/specification-document) (with a CVSS 4.0 approximation); wired into both `parseCvssScore` and `extractCvssScoreFromOsvVulnerability`. Tests: `CvssCalculatorTest` (12 tests) + `VulnerabilityServiceTest.cvssVectorScoreIsParsedToNumericValue` |
| 13 | Cold-cache loader UX (Promise.all blocks SBOM table) | `bd96b79e` — split into three independent `useEffect`s; SBOM table now renders as soon as the SBOM fetch resolves. Already marked resolved by Claudia |
| 14 | `refreshVersionInfo()` dead code | Removed in `03fccab2`. (Claudia later said "fine to keep" once tested — went ahead and removed it since no callers exist) |

### Bonus (CodeRabbit nitpick) — cache stampede

Both `ThesisManagementVersionService.getVersionInfo` and `VulnerabilityService.getVulnerabilities` now use double-checked synchronization, so a thundering herd of admins at cache-expiry no longer triggers parallel OSV / GitHub calls.

---

**Test deltas in this round**
- `CvssCalculatorTest` — 12 new tests
- `VulnerabilityServiceTest` — 5 new (Maven `.RELEASE`/`.Final` normalization, multi-event ranges, vector-string CVSS parsing)
- `DependencyOverviewControllerTest` — refresh asserted as POST; added 405 assertion for GET
- `ThesisManagementVersionServiceTest` — `refreshVersionInfo()` calls replaced; dropped the now-impossible bypass test

**Local verification**: server 862 tests pass, client tsc + eslint clean, vitest 84 tests pass.

server/src/main/java/de/tum/cit/aet/thesis/dependency/service/VulnerabilityService.java
Claudia-Anthropica: @krusche Nice — this is the real fix. `parseCvssScore()` now falls back to `CvssCalculator.scoreFromVector()`, so the vector string actually gets turned into a base score, `cvssScore` is populated, and `mapCvssScoreToSeverity()` runs on real data instead of the heuristic. The new `cvssVectorScoreIsParsedToNumericValue` test even uses the genuine OSV shape (`CVSS_V3` + full vector string), and `CvssCalculatorTest` asserts a bare `"7.5"` returns null — exactly backwards from before. The CVSS 3.1 formula and roundup look correct too. Resolving.

Claudia-Anthropica (approved):
@krusche All my feedback is addressed. The CVSS path is now real — `CvssCalculator` parses the OSV vector string into a base score so `mapCvssScoreToSeverity()` actually runs, backed by a proper test using the genuine OSV data shape. You also fixed the multi-interval range handling and the Maven version-qualifier over-reporting along the way. Nice work.

server/src/main/java/de/tum/cit/aet/thesis/dependency/service/VulnerabilityService.java
```diff
@@ -0,0 +1,693 @@
...
+					return true;
+				}
+			}
+			return false;
+		} catch (SemverException e) {
+			log.warn("Failed to parse version '{}': {}", componentVersion, e.getMessage());
+			return true;
+		}
+	}
+
```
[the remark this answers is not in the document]
krusche: Fixed in 03fccab2. `VulnerabilityService.parseVersionLenient` now strips Maven qualifiers (`.RELEASE`/`.Final`/`.GA`/`.SP*`/`.CR*`) and collapses 4-part versions before NPM Semver parsing; on persistent failure returns `false` (conservative, no over-reporting). Test: `VulnerabilityServiceTest.mavenQualifierVersionDoesNotOverReportCves` and `mavenFinalQualifierIsNormalized`.

server/src/main/java/de/tum/cit/aet/thesis/dependency/service/VulnerabilityService.java
```diff
@@ -0,0 +1,693 @@
...
+							case "MEDIUM" -> mediumSeverityCount++;
+							case "LOW" -> lowSeverityCount++;
+							default -> {
+								// UNKNOWN severity is counted in total but not in specific categories
+							}
+						}
+					}
+				}
+			} catch (Exception e) {
+				log.error("Failed to query OSV API for batch starting at index {}: {}", batchStartIndex, e.getMessage());
```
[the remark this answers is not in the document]
krusche: Fixed in 03fccab2. Switched to `log.error("Failed to query OSV API for batch starting at index {}", batchStartIndex, e)` so the stack trace is preserved. Same change applied to `sendBatchRequestToOsvApi`.

server/src/main/java/de/tum/cit/aet/thesis/dependency/controller/DependencyOverviewController.java
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * REST controller for the admin dependency overview page. Exposes the bundled SBOMs,
+ * OSV-derived vulnerability information, version-update status, and a manual trigger
+ * for the weekly vulnerability email.
+ */
+@RestController
+@RequestMapping("/v2/admin/dependencies")
+@PreAuthorize("hasRole('admin')")
+public class DependencyOverviewController {
```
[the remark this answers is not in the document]
krusche: Not a bug — the project has `server.servlet.context-path: /api` in `application.yml`, so the controller-relative `@RequestMapping("/v2/admin/dependencies")` lands at the full URL `/api/v2/admin/dependencies` that the PR description documents. Both are correct.

server/src/main/java/de/tum/cit/aet/thesis/dependency/controller/DependencyOverviewController.java
```diff
@@ -0,0 +1,153 @@
...
+	 * @return 200 with the freshly fetched vulnerability data, or 404 if no SBOM is bundled
+	 */
+	@GetMapping("/vulnerabilities/refresh")
+	public ResponseEntity<ComponentVulnerabilitiesDTO> refreshVulnerabilities() {
+		log.info("REST request to refresh SBOM vulnerabilities");
+		if (!sbomService.isSbomAvailable()) {
+			return ResponseEntity.notFound().build();
+		}
+		return ResponseEntity.ok(vulnerabilityService.refreshVulnerabilities());
+	}
```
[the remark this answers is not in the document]
krusche: Fixed in 03fccab2. Changed to `@PostMapping("/vulnerabilities/refresh")`; client request updated to POST. Test: `DependencyOverviewControllerTest.refreshIsPostAndReturnsCounts` and `refreshRejectsGetMethod` (expects 405 on GET).

server/src/main/java/de/tum/cit/aet/thesis/dependency/controller/DependencyOverviewController.java
```diff
@@ -0,0 +1,153 @@
...
+	 */
+	@GetMapping
+	public ResponseEntity<CombinedSbomDTO> getCombinedSbom() {
+		log.debug("REST request to get combined SBOM");
+		if (!sbomService.isSbomAvailable()) {
+			return ResponseEntity.notFound().build();
+		}
+		return ResponseEntity.ok(sbomService.getCombinedSbom());
+	}
+
```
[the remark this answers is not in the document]
krusche: Fixed in bd96b79e. `DependencyOverviewControllerTest` adds 11 tests covering: (1) admin vs supervisor vs unauthenticated gating on every endpoint, (2) 404 paths via SBOM fixture overrides, (3) `{"sent": true}` shape on the send-email endpoint.

server/src/main/java/de/tum/cit/aet/thesis/dependency/service/VulnerabilityScanMailSender.java
```diff
@@ -0,0 +1,108 @@
...
+			message.setContent(html, "text/html; charset=utf-8");
+
+			if (mailConfig.isEnabled()) {
+				mailSender.send(message);
+				log.info("Sent vulnerability scan email to {}", recipientEmail);
+			} else {
+				log.debug("Vulnerability scan email (postfix disabled)\n{}", MailLogger.getTextFromMimeMessage(message));
+			}
+		} catch (Exception e) {
+			log.warn("Failed to send vulnerability scan email to {}", recipientEmail, e);
```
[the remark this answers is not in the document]
krusche: Fixed in 03fccab2. `VulnerabilityScanMailSender.send()` now returns `boolean`; `VulnerabilityScanScheduleService.sendVulnerabilityScanEmail()` inspects it and returns `false` (→ controller 400 with `{"sent": false\}`) when delivery fails. Exceptions are still logged inside `send()` so the scheduler keeps running.

server/src/main/java/de/tum/cit/aet/thesis/dependency/service/SbomService.java
```diff
@@ -0,0 +1,229 @@
...
+	public SbomDTO getClientSbom() {
+		return loadSbomCached(clientSbomPath);
+	}
+
+	/**
+	 * @return true if at least one SBOM file is bundled in the executable
+	 */
+	public boolean isSbomAvailable() {
+		return isResourceAvailable(serverSbomPath) || isResourceAvailable(clientSbomPath);
+	}
```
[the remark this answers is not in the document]
krusche: Fixed in a54dbece. `isSbomAvailable()` now requires successful parse, not just file presence: `return getServerSbom() != null || getClientSbom() != null;`. Test: `SbomServiceTest.isSbomAvailableFalseWhenFilePresentButUnparseable` with a deliberately invalid JSON fixture.

server/src/main/java/de/tum/cit/aet/thesis/dependency/service/VulnerabilityService.java
```diff
@@ -0,0 +1,794 @@
...
+				}
+				currentIntroduced = null;
+				haveOpenInterval = false;
+				continue;
+			}
+
+			if (event.limit() != null) {
+				Semver limit = parseVersionLenient(event.limit());
+				if (intervalMatches(installedVersion, currentIntroduced, limit, /*inclusiveUpperBound*/ true)) {
+					return true;
```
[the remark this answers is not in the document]
krusche: Fixed in a54dbece. `rangeContainsVersion` now guards both the `fixed` and `limit` branches with a null check on the parsed Semver, skipping the interval rather than letting `intervalMatches` receive `upper=null` and treat it as unbounded. Test: `VulnerabilityServiceTest.unparsableFixedBoundIsTreatedAsNotMatching`.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:
@krusche on it! Taking a look now :eyes:

Claudia-Anthropica (approved):
@krusche This round nails the last two CodeRabbit findings cleanly — `isSbomAvailable()` now keys off the actual parse result so a corrupt SBOM no longer masquerades as available, and `rangeContainsVersion` correctly skips intervals whose closing bound won't parse instead of treating them as unbounded (which was a real false-positive source). Both come with realistic tests that exercise the full path. Nothing new on my end. Good to go.

server/src/main/resources/application.yml
```diff
@@ -98,3 +98,13 @@ thesis-management:
     days-between-exports: ${DATA_EXPORT_COOLDOWN_DAYS:7}
   storage:
     upload-location: ${UPLOAD_FOLDER:uploads}
+  dependency-scan:
+    cron: ${VULNERABILITY_SCAN_CRON:0 0 8 * * MON}
+    # Defaults to the support address used in the About page so the weekly scan
+    # works without any deployment-level config. Override via env var if a different
+    # mailbox should receive the report.
+    admin-email: ${VULNERABILITY_SCAN_ADMIN_EMAIL:thesis-management-support.aet@xcit.tum.de}
```
Claudia-Anthropica: @krusche [low] Heads up — this defaults `admin-email` to the real TUM support address, but the PR description documents a blank default (and notes a blank value makes the cron skip). So any deployment that doesn't set `VULNERABILITY_SCAN_ADMIN_EMAIL` will email `thesis-management-support.aet@xcit.tum.de` weekly instead of skipping — other chairs/forks would silently send their reports to TUM, and their own admin would never get one. If that's intended for the primary deployment, fine; otherwise a blank default (the code already handles blank → skip) matches the docs and avoids the surprise.

In `server/src/main/resources/application.yml`, the `thesis-management.dependency-scan.admin-email` property defaults to a hardcoded TUM support address, so any deployment that doesn't set `VULNERABILITY_SCAN_ADMIN_EMAIL` emails that address weekly. Change the default to empty (`${VULNERABILITY_SCAN_ADMIN_EMAIL:}`) so unconfigured deployments skip the scan, matching the documented behavior.

</details>

server/src/main/java/de/tum/cit/aet/thesis/dependency/service/VulnerabilityService.java
```diff
@@ -0,0 +1,797 @@
...
+			return true;
+		}
+
+		Semver installedVersion = parseVersionLenient(componentVersion);
+		if (installedVersion == null) {
+			// Unable to parse — fall back to lastKnownAffectedVersionRange comparison where
+			// possible, otherwise treat as not affected to avoid spamming admins with false
+			// positives on every Maven dependency that has a qualified version like
+			// "3.1.5.RELEASE" / "4.2.12.Final".
+			log.debug("Could not parse version '{}' for {}; trying range-string fallback",
```
Claudia-Anthropica: @krusche [nit] This log line (and the comment just above it) says it'll fall back to `lastKnownAffectedVersionRange` / try a range-string fallback, but the branch just `return false` — no fallback actually runs. Worth fixing the wording so someone debugging an empty scan doesn't go looking for a phantom code path.

In `server/src/main/java/de/tum/cit/aet/thesis/dependency/service/VulnerabilityService.java`, the `installedVersion == null` branch in `isVersionAffectedByVulnerability` logs "trying range-string fallback" and the comment mentions falling back to lastKnownAffectedVersionRange, but it just returns false. Update the comment and log message to state that an unparseable version is treated as not-affected, with no fallback.

</details>

Claudia-Anthropica (approved):
@krusche Re-reviewed the whole feature end-to-end. It's in good shape — the CVSS path is real (CvssCalculator follows the v3.1 spec), the Maven version normalization, decoupled client fetches, parseability-aware isSbomAvailable, and multi-interval range handling from earlier rounds all hold up. Two small things I hadn't caught before, both non-blocking: the hardcoded admin-email default diverges from the documented blank default (inline), and a stale log/comment in VulnerabilityService. Approving.

server/src/main/resources/application.yml
krusche: Fixed in 57f84ef9. Dropped the new `VULNERABILITY_SCAN_ADMIN_EMAIL` env var entirely; `VulnerabilityScanScheduleService` now uses the existing `mailConfig.getSender().getAddress()` (i.e. `MAIL_SENDER`) as the scan-report recipient. The TUM deployment already sets `MAIL_SENDER` to the support address via repo vars, so the primary deployment keeps working with zero extra config. Forks that set their own `MAIL_SENDER` for outbound mail automatically get the scan to their own mailbox — no risk of accidentally emailing TUM. Blank sender → cron logs a warning and skips, matching the documented behavior.

server/src/main/java/de/tum/cit/aet/thesis/dependency/service/VulnerabilityService.java
krusche: Fixed in ae8f479e. The comment now reads "treat as not affected — there's nothing to fall back to in this path" and the log message says "treating as not affected" so a future debugger sees the actual behavior.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:
@krusche roger that — review incoming!

Claudia-Anthropica (approved):
@krusche Both of my remaining points are handled cleanly. Dropping the dedicated `VULNERABILITY_SCAN_ADMIN_EMAIL` and reading `mailConfig.getSender()` instead is a nicer fix than just documenting the TUM default — forks now email whatever `MAIL_SENDER` they already configure, with a sensible skip+warn when it's blank. The version-parse comment/log no longer references a phantom range-string fallback. Everything else still looks solid and CI is green. Approving.

PR #1149, opened 2026-06-28
Add initial AI feedback module for PDF thesis artifacts

bensofficial:
## Summary

Adds an experimental, feature-flagged AI feedback module that reviews uploaded thesis PDFs (initially proposals) by chunking the document into per-category LLM passes and merging the findings into a single structured result.

- New `POST /v2/ai-review/review-proposal` endpoint (multipart upload, restricted to `admin`/`advisor`/`supervisor`) returning a `ReviewResultDTO` with findings (severity, title, category, description, page/section/quote locations).
- `ReviewService` runs each `ReviewCategory` through `LlmReviewer` and merges intermediate results via a final `Prompts.MERGER` call against Spring AI's `ChatClient`. `PdfService` extracts per-page text and renders per-page images so the LLM gets both modalities.
- Gated by a Spring `Condition` (`AIFeaturesEnabled` → `thesis-management.ai.enabled`). When off, the controller and services aren't registered and the endpoint returns 404. Off by default in `application.yml`; enabled in `application-dev.yml` for local dev only.
- Wires Spring AI's OpenAI starter (`spring-ai-starter-model-openai`) in `server/build.gradle`. Configurable via `OPENAI_API_KEY` / `OPENAI_BASE_URL` / `OPENAI_CHAT_MODEL` — defaults point at the chair's TUM GPU endpoint.
- Documents the new flags in `docs/CONFIGURATION.md` and the experimental status / dev setup in `docs/DEVELOPMENT.md`.
- Adds unit tests for `ReviewController` (with `@TestPropertySource` flipping the flag on), `ReviewService`, `PdfService`, and `Prompts`, plus fixture PDFs/images under `server/src/test/resources/pdfs`.

> **Status:** work in progress / preview. The Azure provider path explicitly returns 400, and the controller still re-uploads the file rather than reusing the thesis service's already-stored artifact (TODO in the code). Treat as a foundation, not a finished feature.

## Test plan

- [ ] `cd server && ./gradlew test` — including the new `feedback` package tests
- [ ] `cd server && ./gradlew spotlessApply` — no formatting drift
- [ ] With the flag off (default), confirm `POST /v2/ai-review/review-proposal` returns 404
- [ ] With the flag on and a reachable LLM endpoint, upload `proposal-template.pdf` and verify a populated `ReviewResultDTO` comes back

## Summary by CodeRabbit

* **New Features**
  * Added an experimental AI feedback flow for uploaded PDFs via `POST /v2/ai-review/review-proposal`, returning structured review results (category, summary, findings with per-page locations/images).
  * Added PDF extraction to support per-page text and rendered page images for review.
* **Bug Fixes**
  * AI review is now opt-in via a feature flag; when disabled, the endpoint is not available, and unsupported provider categories are rejected.
* **Documentation**
  * Documented new AI environment variables and added an “AI Feedback (Experimental)” configuration/feature-flag guide (including dev/test behavior).
* **Tests**
  * Added integration/unit tests for the new endpoint, PDF extraction, and prompt/category mapping (including image snapshot verification).
* **Chores**
  * Updated the server SBOM input fingerprint.


server/build.gradle
```diff
@@ -67,6 +67,10 @@ dependencies {
 	implementation "org.springframework.boot:spring-boot-starter-thymeleaf"
 	implementation "org.springframework.boot:spring-boot-starter-actuator"
 
+	implementation platform("org.springframework.ai:spring-ai-bom:2.0.0")
+	implementation "org.springframework.ai:spring-ai-pdf-document-reader"
+	implementation 'org.springframework.ai:spring-ai-starter-model-openai'
```
Claudia-Anthropica, line 72: @bensofficial [medium] These new runtime dependencies are not reflected in `server/sbom/bom.json` or `server/sbom/.input-hash`. The admin dependency page and vulnerability scan bundle the committed SBOM from `processResources`, so Spring AI and its transitive dependencies will be invisible there until the SBOM is regenerated and committed.

In `server/build.gradle`, new Spring AI dependencies were added but the committed server SBOM files were not updated. Run the server CycloneDX generation task and commit the resulting `server/sbom/bom.json` and `server/sbom/.input-hash` changes.

</details>

server/src/main/java/de/tum/cit/aet/thesis/feedback/controller/ReviewController.java
```diff
@@ -0,0 +1,55 @@
...
+	 *
+	 * @param request multipart payload containing the proposal file and the provider category
+	 * @return the merged review result produced by the LLM pipeline
+	 */
+	@PostMapping(value = "review-proposal", consumes = MediaType.MULTIPART_FORM_DATA_VALUE)
+	@PreAuthorize("hasAnyRole('admin', 'advisor', 'supervisor')")
+	public ResponseEntity<ReviewResultDTO> reviewProposal(@ModelAttribute ReviewRequestDTO request) {
+		// TODO: Use already uploaded file from the thesis service instead of uploading it again
+
+		if (request.providerCategory().equals(ProviderCategory.AZURE)) {
```
Claudia-Anthropica, line 50: @bensofficial [medium] `providerCategory()` can be null when the multipart field is omitted, and `file()` can be null as well; those cases currently turn into NPE/500 responses instead of a client-side 400. Validate both fields before using them, or bind them as required request parts so Spring rejects malformed requests cleanly.

In `server/src/main/java/de/tum/cit/aet/thesis/feedback/controller/ReviewController.java`, missing multipart fields can produce null values and then a 500 response. Validate `request.providerCategory()` and `request.file()` before use, returning the existing invalid-parameters/400 behavior for missing values.

</details>

server/src/main/java/de/tum/cit/aet/thesis/feedback/service/PdfService.java
```diff
@@ -0,0 +1,89 @@
...
+	 * @return one PNG-encoded {@link Media} per page in document order
+	 */
+	public List<Media> extractImagesFromPdf(MultipartFile file) {
+		log.debug("Extracting images from PDF file: {}", file.getOriginalFilename());
+		List<Media> images = new ArrayList<>();
+
+		try (PDDocument document = Loader.loadPDF(file.getBytes())) {
+			PDFRenderer renderer = new PDFRenderer(document);
+
+			for (int page = 0; page < document.getNumberOfPages(); page++) {
```
Claudia-Anthropica, line 76: @bensofficial [medium] The 25 MB upload cap does not cap page count, and this loop renders every page at 300 DPI before the review sends those images through all category passes. A small but long PDF can still allocate a large amount of memory and trigger a very expensive LLM request; reject proposals above a small page limit before rendering.

In `server/src/main/java/de/tum/cit/aet/thesis/feedback/service/PdfService.java`, PDF rendering has no page-count guard. Add a configurable or fixed proposal page limit before rendering/extracting all pages and return a controlled client error when the PDF exceeds it.

</details>

server/src/test/java/de/tum/cit/aet/thesis/feedback/service/PdfServiceTest.java
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	@Test
+	public void testExtractTextFromSinglePagePdf() {
+		MockMultipartFile multipartFile = new MockMultipartFile("hello-world", "hello-world.pdf", "application/pdf", helloWorldContent);
+
+		List<String> actual = pdfService.extractTextFromPdf(multipartFile);
+
+		assertEquals(1, actual.size(), "Expected exactly one page of text to be extracted");
+		assertEquals(List.of("""
```
Claudia-Anthropica: @bensofficial [medium] This exact text snapshot fails in CI and locally because the extracted PDF text has different leading whitespace than the literal here. Normalize the extracted text or assert stable substrings like the multi-page test does, otherwise `./gradlew test` stays red.

In `server/src/test/java/de/tum/cit/aet/thesis/feedback/service/PdfServiceTest.java`, `testExtractTextFromSinglePagePdf` fails because it compares the full extracted PDF text including unstable whitespace. Replace the exact full-string assertion with normalized text comparison or focused stable-content assertions.

</details>

server/src/test/java/de/tum/cit/aet/thesis/feedback/service/ReviewServiceTest.java
```diff
@@ -0,0 +1,129 @@
...
+				## [LOW] Clear writing style
+				Category: writing-style
+				Description: The writing style is clear.
+				Locations:
+				- Page 2, Section: Methodology, Quote: "The methodology section is well written."
+				""";
+
+		// buildMergePrompt iterates an unordered Map, so assert each section is present
+		// independently rather than expecting a fixed ordering.
+		assertThat(mergePrompt).contains(structureSection);
```
Claudia-Anthropica: @bensofficial [medium] This assertion currently fails because `buildMergePrompt()` emits two spaces before each location item, while the expected text starts the dash at column zero. Update the expected sections to match the generated prompt or make the assertion less whitespace-sensitive.

In `server/src/test/java/de/tum/cit/aet/thesis/feedback/service/ReviewServiceTest.java`, `testBuildMergePrompt` expects location lines without the indentation that `buildMergePrompt()` actually emits. Fix the expected strings or normalize indentation before asserting so the test passes.

</details>

Claudia-Anthropica (changes requested):
@bensofficial The AI review pipeline shape is understandable, but this head currently has two failing server tests and the new Spring AI dependencies leave the bundled SBOM stale. I also left comments on request validation and PDF rendering bounds; please fix these before merge.

server/src/main/java/de/tum/cit/aet/thesis/feedback/controller/ReviewController.java
bensofficial: -> Follow-up PR

server/src/main/java/de/tum/cit/aet/thesis/feedback/service/PdfService.java
bensofficial: -> Follow-up PR

server/src/main/java/de/tum/cit/aet/thesis/feedback/controller/ReviewController.java
Claudia-Anthropica: @bensofficial I would keep this in this PR: the endpoint is added here and malformed multipart requests still hit a null dereference before returning a controlled 400. Please add the validation here rather than deferring it.

server/src/main/java/de/tum/cit/aet/thesis/feedback/service/PdfService.java
Claudia-Anthropica: @bensofficial I would also keep the page-count guard in this PR. This endpoint renders every uploaded page at 300 DPI before review, so a small long PDF can still cause large memory use and an expensive LLM request; please cap proposals before rendering instead of deferring that.

Claudia-Anthropica (changes requested):
@bensofficial The SBOM and the failing feedback tests are fixed now; the targeted feedback test package passes locally. I am still requesting changes because the request validation and PDF page-count guard are still missing in this PR, and I replied on those two threads.

server/src/main/java/de/tum/cit/aet/thesis/feedback/service/PdfService.java
Claudia-Anthropica: @bensofficial This still needs a tighter cap. `MAX_PAGES = 120` prevents extremely large PDFs, but for a proposal endpoint it still allows rendering and sending 120 pages of 300 DPI images through every category pass. Please cap this to a genuinely proposal-sized limit before rendering.

Claudia-Anthropica (changes requested):
@bensofficial The validation, SBOM, and previously failing feedback tests look good now; I ran the targeted feedback package locally and it passes. I’m still requesting changes because the PDF page-count guard is set to 120 pages, which is too loose for this proposal endpoint; I reopened and replied on that thread.

server/src/main/java/de/tum/cit/aet/thesis/feedback/service/PdfService.java
bensofficial: The idea is that this endpoint is used not only for proposals but also for the thesis. Right now, at this early prototype stage, the prompts are just for proposals, true, but in general, it should support both. And for the final thesis, 120 pages are already a rather low limit.

Claudia-Anthropica: @bensofficial Good point, if this is intended to cover thesis-sized uploads too, I’m fine treating the 120-page cap as the current product constraint rather than blocking on a proposal-only limit.

Claudia-Anthropica (approved):
@bensofficial All requested feedback is settled now; I’m accepting the 120-page cap as the current constraint for thesis-sized uploads. Build/test/security checks are passing, with only the dev-container deploy job still pending, so approving.

server/src/main/java/de/tum/cit/aet/thesis/feedback/service/reviewer/LlmReviewer.java
```diff
@@ -0,0 +1,62 @@
...
+	 * @param images  per-page rendered images in document order
+	 * @return the intermediate review result parsed from the LLM response
+	 */
+	public IntermediateReviewResult review(List<String> pages, List<Media> images) {
+		String pagesText = IntStream.range(0, pages.size())
+				.mapToObj(index -> "=== PAGE " + (index + 1) + " ===\n" + pages.get(index))
+				.collect(Collectors.joining("\n\n"));
+
+		return chatClient.prompt().system(systemMessage -> systemMessage.text(sharedPrompt + "\n\n" + taskPrompt + "\n\n" + guidelinesPrompt))
+				.user(userMessage -> userMessage.text(pagesText).media(images.toArray(new Media[0]))).call().entity(IntermediateReviewResult.class);
```
Claudia-Anthropica: @bensofficial [medium] The category reviewer sends the uploaded PDF text as the user message without telling the model that this is untrusted document content. A student can put prompt-like text in the PDF, such as instructions to ignore the rubric or return no findings, and that competes with the reviewer prompt before the merger ever sees the result. Add a data-only instruction and fence the page text so the first pass treats the PDF text and page images strictly as evidence, not instructions.

In `server/src/main/java/de/tum/cit/aet/thesis/feedback/service/reviewer/LlmReviewer.java`, the uploaded PDF text is sent directly as the user message without guardrails telling the model to treat it as untrusted document content. Fix this by adding explicit system instructions that page text and page images are data from a student upload, and wrap the generated `pagesText` in a clear data fence before sending it to the model.

</details>

Claudia-Anthropica (changes requested):
@bensofficial The latest fixes address the previous threads and the targeted feedback tests pass locally. I found one remaining prompt-injection gap in the first reviewer pass; see the inline comment.

server/src/main/java/de/tum/cit/aet/thesis/feedback/service/reviewer/LlmReviewer.java
bensofficial: Fixed in 173b8ee8

Claudia-Anthropica: @bensofficial Looks good now, thanks.

Claudia-Anthropica (approved):
@bensofficial The latest fix addresses the LlmReviewer prompt-injection thread, and the focused test passes locally. The build/test/security checks are passing; only the dev-container deploy job is still pending, so I’m approving.

Claudia-Anthropica (approved):
@bensofficial The dependency/SBOM refresh looks fine; I rechecked the feedback module and the calendar paths affected by the ical4j bump. Focused local tests passed, and GitHub build/test/e2e are green with only deploy and JavaScript/TypeScript CodeQL still pending, so approving.

PR #1157, opened 2026-06-30
Preserve table pagination on back-nav and enable new-tab row clicks

bensofficial:
## Summary
- **New-tab row clicks**: Cmd/Ctrl/Shift+click and middle-click on rows in the theses, topics, public presentations, and dashboard task tables now open the target in a new tab. Plain click still navigates in the current tab. Removes the partial `<Link>` on the `TopicsTable` title cell that previously caused double-fire on Cmd+click.
- **Pagination/filters/sort persistence**: `ThesesProvider`, `TopicsProvider`, and `ApplicationsProvider` gain an opt-in `persistState` prop that mirrors page/search/filters/sort to URL search params (using `replace: true` so the URL doesn't pollute history). Enabled on `/theses`, `/topics`, and `/applications`. Browser back from a detail page now restores the previous list view.
- **Scoping**: `persistState` defaults to `false`, so call sites that share a route with other URL-owning state are unaffected (Dashboard, LandingPage, TopicPage, IntervieweesList, NotificationSettings, ThesisOverviewPage).

## Test plan
- [x] On `/theses`, page to page 2, open a thesis (plain click) → browser back → still on page 2.
- [x] On `/theses`, Cmd/Ctrl/Shift+click a row → opens detail in a new tab; current tab unchanged.
- [x] On `/theses`, middle-click a row → opens detail in a new tab.
- [x] Apply a state/type filter or change sort → URL reflects it (`?states=...&sortBy=...`); reload preserves filters; navigating away via the "Browse Theses" nav link resets to defaults.
- [x] Repeat the above on `/topics`.
- [x] On `/applications`, paginate the sidebar, select an application, then browser-back → previous page restored.
- [x] On the dashboard My Tasks section and the public presentations table, Cmd+click / middle-click → opens in new tab.
- [x] Regression: dashboard, landing page, and topic-page tables behave as before (no URL params change, no double navigation).

## Summary by CodeRabbit

* **New Features**
  * Added URL persistence for list state (filters, sort, and page) for applications, topics, and theses.
  * Updated table row navigation to support opening details in a new tab via modifier keys and middle-click.
  * Enhanced table components to provide richer row-click handling (including access to the click event and customizable row attributes).
* **Bug Fixes**
  * Improved consistency of multi-tab navigation behavior across list pages and tables.

### DIFF
--- a/client/src/core/application/pages/ReviewApplicationPage/ReviewApplicationPage.tsx
+++ b/client/src/core/application/pages/ReviewApplicationPage/ReviewApplicationPage.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from 'react'
-import { useNavigate, useParams } from 'react-router'
+import { useLocation, useNavigate, useParams } from 'react-router'
 import type { IApplication } from '@/core/application/requests/responses/application'
 import { ApplicationState } from '@/core/application/requests/responses/application'
 import { doRequest } from '@/core/requests/request'
@@ -12,6 +12,7 @@ import ApplicationModal from '@/core/application/components/ApplicationModal/App
 
 const ReviewApplicationPage = () => {
   const navigate = useNavigate()
+  const location = useLocation()
   const { applicationId } = useParams<{ applicationId: string }>()
 
   usePageTitle('Review Applications')
@@ -49,7 +50,7 @@ const ReviewApplicationPage = () => {
         <ApplicationModal
           application={application}
           onClose={() => {
-            void navigate('/applications', { replace: true })
+            void navigate({ pathname: '/applications', search: location.search }, { replace: true })
 
             setApplication(undefined)
           }}
@@ -65,9 +66,13 @@ const ReviewApplicationPage = () => {
             selected={application}
             isSmallScreen={isSmallScreen}
             onSelect={(newApplication) => {
-              void navigate(`/applications/${newApplication.applicationId}`, {
-                replace: true,
-              })
+              void navigate(
+                {
+                  pathname: `/applications/${newApplication.applicationId}`,
+                  search: location.search,
+                },
+                { replace: true },
+              )
 
               setApplication(newApplication)
             }}
@@ -81,7 +86,10 @@ const ReviewApplicationPage = () => {
                 onChange={setApplication}
                 onDelete={() => {
                   setApplication(undefined)
-                  void navigate('/applications', { replace: true })
+                  void navigate(
+                    { pathname: '/applications', search: location.search },
+                    { replace: true },
+                  )
                 }}
               />
             ) : (
--- a/client/src/core/application/pages/ReviewApplicationPage/ReviewApplicationPage.tsx
+++ b/client/src/core/application/pages/ReviewApplicationPage/ReviewApplicationPage.tsx
@@ -43,6 +43,7 @@ const ReviewApplicationPage = () => {
       limit={10}
       defaultStates={[ApplicationState.NOT_ASSESSED]}
       showOnlyAssignedTopics={true}
+      persistState
     >
       {isSmallScreen && (
         <ApplicationModal
--- a/client/src/core/application/providers/ApplicationsProvider/ApplicationsProvider.tsx
+++ b/client/src/core/application/providers/ApplicationsProvider/ApplicationsProvider.tsx
@@ -124,6 +124,9 @@ const ApplicationsProvider = (props: PropsWithChildren<IApplicationsProviderProp
   const filterStatesKey = adjustedFilters.states?.join(',')
   const filterTopicsKey = adjustedFilters.topics?.join(',')
   const filterTypesKey = adjustedFilters.types?.join(',')
+  // URL sync uses the raw user-selected topics so the implicit
+  // `showOnlyAssignedTopics` expansion is not leaked into `?topics=` params.
+  const urlTopicsKey = filters.topics?.join(',')
   const topicsLoaded = !!topics
 
   const didMountRef = useRef(false)
@@ -219,14 +222,14 @@ const ApplicationsProvider = (props: PropsWithChildren<IApplicationsProviderProp
     setOrDelete('page', page > 0 ? String(page) : undefined)
     setOrDelete('search', filters.search)
     setOrDelete('states', filterStatesKey)
-    setOrDelete('topics', filterTopicsKey)
+    setOrDelete('topics', urlTopicsKey)
     setOrDelete('types', filterTypesKey)
     setOrDelete('sortBy', sort.column !== DEFAULT_SORT.column ? sort.column : undefined)
     setOrDelete('sortOrder', sort.direction !== DEFAULT_SORT.direction ? sort.direction : undefined)
 
     setSearchParams(params, { replace: true })
     // eslint-disable-next-line @eslint-react/exhaustive-deps -- searchParams/setSearchParams change on every navigation; only sync URL when actual state changes
-  }, [persistState, page, filters.search, filterStatesKey, filterTopicsKey, filterTypesKey, sort])
+  }, [persistState, page, filters.search, filterStatesKey, urlTopicsKey, filterTypesKey, sort])
 
   const fetchApplication = async (applicationId: string): Promise<IApplication | null> => {
     return new Promise((resolve) => {
--- a/client/src/core/application/providers/ApplicationsProvider/ApplicationsProvider.tsx
+++ b/client/src/core/application/providers/ApplicationsProvider/ApplicationsProvider.tsx
@@ -34,6 +34,8 @@ interface IApplicationsProviderProps {
 }
 
 const DEFAULT_SORT: IApplicationsSort = { column: 'createdAt', direction: 'desc' }
+const SORT_COLUMNS: IApplicationsSort['column'][] = ['createdAt', 'updatedAt']
+const SORT_DIRECTIONS: IApplicationsSort['direction'][] = ['asc', 'desc']
 
 const ApplicationsProvider = (props: PropsWithChildren<IApplicationsProviderProps>) => {
   const {
@@ -86,11 +88,15 @@ const ApplicationsProvider = (props: PropsWithChildren<IApplicationsProviderProp
   })
   const [sort, setSort] = useState<IApplicationsSort>(() => {
     if (!persistState) return DEFAULT_SORT
-    const column = searchParams.get('sortBy') as IApplicationsSort['column'] | null
-    const direction = searchParams.get('sortOrder') as IApplicationsSort['direction'] | null
+    const column = searchParams.get('sortBy')
+    const direction = searchParams.get('sortOrder')
     return {
-      column: column ?? DEFAULT_SORT.column,
-      direction: direction ?? DEFAULT_SORT.direction,
+      column: SORT_COLUMNS.includes(column as IApplicationsSort['column'])
+        ? (column as IApplicationsSort['column'])
+        : DEFAULT_SORT.column,
+      direction: SORT_DIRECTIONS.includes(direction as IApplicationsSort['direction'])
+        ? (direction as IApplicationsSort['direction'])
+        : DEFAULT_SORT.direction,
     }
   })
 
@@ -120,7 +126,12 @@ const ApplicationsProvider = (props: PropsWithChildren<IApplicationsProviderProp
   const filterTypesKey = adjustedFilters.types?.join(',')
   const topicsLoaded = !!topics
 
+  const didMountRef = useRef(false)
   useEffect(() => {
+    if (!didMountRef.current) {
+      didMountRef.current = true
+      return
+    }
     setPage(0)
   }, [sort, adjustedFilters])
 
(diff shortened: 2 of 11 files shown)


client/src/core/application/pages/ReviewApplicationPage/ReviewApplicationPage.tsx
```diff
@@ -43,6 +43,7 @@ const ReviewApplicationPage = () => {
       limit={10}
       defaultStates={[ApplicationState.NOT_ASSESSED]}
       showOnlyAssignedTopics={true}
+      persistState
```
Claudia-Anthropica, line 47: @bensofficial [medium] Turning on `persistState` here does not keep the URL state once an application is selected. The `navigate(`/applications/${newApplication.applicationId}`, { replace: true })` call below drops `location.search`, so paging to page 2 and selecting an item leaves the detail URL without `?page=1`; reload/share/back-to-this-entry restores the default page instead. Preserve the current search params when navigating between `/applications` and `/applications/:id`.

In `client/src/core/application/pages/ReviewApplicationPage/ReviewApplicationPage.tsx`, the route enables `persistState` but navigation to `/applications/:applicationId` drops the current query string. Fix the application selection, close, and delete navigations so they preserve the provider-owned search params when moving within the applications route.

</details>

client/src/core/application/providers/ApplicationsProvider/ApplicationsProvider.tsx
```diff
@@ -38,24 +45,53 @@ const ApplicationsProvider = (props: PropsWithChildren<IApplicationsProviderProp
...
 
   const user = useLoggedInUser()
   const topics = useAllTopics()
 
+  const [searchParams, setSearchParams] = useSearchParams()
+
   const [applications, setApplications] = useState<PaginationResponse<IApplication>>()
-  const [page, setPage] = useState(0)
+
+  const [page, setPage] = useState(() => {
```
Claudia-Anthropica, line 60: @bensofficial [medium] This URL `page` value gets wiped on mount. The reset effect at line 123 still runs after the first render and calls `setPage(0)` for the initial `sort`/`adjustedFilters`, so opening `/applications?page=2` immediately falls back to page 1 and rewrites the URL. Guard the first run of that reset, or move page resets into the filter/sort setters.

In `client/src/core/application/providers/ApplicationsProvider/ApplicationsProvider.tsx`, the persisted `page` value from `useSearchParams()` is immediately reset by the `useEffect(() => setPage(0), [sort, adjustedFilters])` that runs on mount. Fix this by skipping that effect's initial run or by resetting the page only from the filter/sort setter paths, so `/applications?page=...` hydrates correctly.

</details>

Claudia-Anthropica (changes requested):
@bensofficial The table-state work is mostly in the right shape, but the `/applications` persistence path currently loses or resets the page state. I left the two application-route issues inline.

client/src/core/application/pages/ReviewApplicationPage/ReviewApplicationPage.tsx
bensofficial: Fixed in 5492a2cf — `ReviewApplicationPage` now preserves `location.search` on all three `navigate` calls (modal close, sidebar select, delete), so paging in `/applications?page=2`, selecting an item, and reloading or browser-back keeps the page state.

client/src/core/application/providers/ApplicationsProvider/ApplicationsProvider.tsx
bensofficial: Fixed in 5492a2cf — guarded the `setPage(0)` reset effect with a `didMountRef` so the page hydrated from `searchParams` survives the first render. Subsequent filter/sort changes still reset to page 0 as before.

client/src/core/application/providers/ApplicationsProvider/ApplicationsProvider.tsx
```diff
@@ -38,24 +45,53 @@ const ApplicationsProvider = (props: PropsWithChildren<IApplicationsProviderProp
...
+    if (!persistState) return 0
+    const raw = parseInt(searchParams.get('page') ?? '', 10)
+    return Number.isFinite(raw) && raw >= 0 ? raw : 0
+  })
```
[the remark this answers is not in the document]
bensofficial: Fixed in 5492a2cf — added a `didMountRef` guard to the reset effect so the URL-hydrated page is preserved on mount.

client/src/core/application/providers/ApplicationsProvider/ApplicationsProvider.tsx
```diff
@@ -38,24 +45,53 @@ const ApplicationsProvider = (props: PropsWithChildren<IApplicationsProviderProp
...
-    column: 'createdAt',
-    direction: 'desc',
+  const [sort, setSort] = useState<IApplicationsSort>(() => {
+    if (!persistState) return DEFAULT_SORT
+    const column = searchParams.get('sortBy') as IApplicationsSort['column'] | null
+    const direction = searchParams.get('sortOrder') as IApplicationsSort['direction'] | null
+    return {
+      column: column ?? DEFAULT_SORT.column,
+      direction: direction ?? DEFAULT_SORT.direction,
+    }
```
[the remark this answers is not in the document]
bensofficial: Fixed in 5492a2cf — `sortBy`/`sortOrder` are now validated against `SORT_COLUMNS` / `SORT_DIRECTIONS` whitelists, falling back to `DEFAULT_SORT` for unknown input.

client/src/thesis/providers/ThesesProvider/ThesesProvider.tsx
```diff
@@ -18,20 +19,58 @@ interface IThesesProviderProps {
...
+  })
+
+  const [sort, setSort] = useState<IThesesSort>(() => {
+    if (!persistState) return DEFAULT_SORT
+    const column = searchParams.get('sortBy') as IThesesSort['column'] | null
+    const direction = searchParams.get('sortOrder') as IThesesSort['direction'] | null
+    return {
+      column: column ?? DEFAULT_SORT.column,
+      direction: direction ?? DEFAULT_SORT.direction,
+    }
```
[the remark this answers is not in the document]
bensofficial: Fixed in 5492a2cf — same validation pattern applied to `ThesesProvider`. Used module-level `SORT_COLUMNS` / `SORT_DIRECTIONS` constants and `includes` checks; unknown URL values fall back to `DEFAULT_SORT`.

bensofficial:
Pushed 5492a2cf addressing all three actionable issues:

- **`ApplicationsProvider` page reset on mount** (Claudia + CodeRabbit): added `didMountRef` guard so the URL-hydrated page survives the first render.
- **`ReviewApplicationPage` losing search params on selection** (Claudia): all three `navigate` calls (modal close, sidebar select, delete) now preserve `location.search`.
- **Unvalidated `sortBy`/`sortOrder` casts** (CodeRabbit, both providers): replaced the blind cast with `SORT_COLUMNS` / `SORT_DIRECTIONS` whitelists; unknown URL values fall back to `DEFAULT_SORT`.

Skipping the nitpicks for now:
- Type-only `React` imports — not adopted elsewhere in the codebase, and the project relies on `React` being in scope in many components; one-off changes would just create inconsistency.
- Extracting shared `openOrNavigate` / `useUrlStateSync` helpers — the duplication is 4–6 lines per call site. Per the repo's CLAUDE.md guidance on avoiding premature abstractions for narrow shared logic, I'd rather wait until a third or fourth provider needs URL sync before factoring this out — at that point the contract is clearer.

client/src/core/application/pages/ReviewApplicationPage/ReviewApplicationPage.tsx
Claudia-Anthropica: @bensofficial Looks good now, thanks. The three navigations preserve `location.search`.

client/src/core/application/providers/ApplicationsProvider/ApplicationsProvider.tsx
Claudia-Anthropica: @bensofficial This still looks unresolved. The guard skips only the first render; once `useAllTopics()` resolves, `adjustedFilters` changes from the placeholder assigned-topic scope to the real topic list, the `[sort, adjustedFilters]` effect runs, and `setPage(0)` wipes the URL-restored page. The reset needs to be tied to user-driven filter/sort changes instead of the async derived `adjustedFilters` value.

client/src/core/application/providers/ApplicationsProvider/ApplicationsProvider.tsx
```diff
@@ -156,6 +203,31 @@ const ApplicationsProvider = (props: PropsWithChildren<IApplicationsProviderProp
...
+        params.set(key, value)
+      } else {
+        params.delete(key)
+      }
+    }
+
+    setOrDelete('page', page > 0 ? String(page) : undefined)
+    setOrDelete('search', filters.search)
+    setOrDelete('states', filterStatesKey)
+    setOrDelete('topics', filterTopicsKey)
```
Claudia-Anthropica: @bensofficial [medium] This writes the derived `adjustedFilters.topics` into the URL, not just the user-selected topic filter. On `/applications` with `showOnlyAssignedTopics`, the URL gets populated with `NO_TOPIC` plus every assigned topic as soon as topics load, even when the user did not choose a topic filter; it also makes that hidden assignment scope stale on reload/share. Sync `filters.topics` to the URL and keep the assigned-topic expansion only inside the request params.

In `client/src/core/application/providers/ApplicationsProvider/ApplicationsProvider.tsx`, the URL sync effect writes `filterTopicsKey` from `adjustedFilters`, so implicit assigned-topic constraints become public `topics` query params. Fix this by deriving the persisted `topics` param from the raw user filter state (`filters.topics`) and keeping the `showOnlyAssignedTopics` expansion only in the API request filters.

</details>

Claudia-Anthropica (changes requested):
@bensofficial The navigation preservation is fixed, but the application pagination issue is still open: `adjustedFilters` changes again after `useAllTopics()` resolves and resets the restored page. I also left one new inline note because the URL sync is persisting the implicit assigned-topic scope as `topics=...`.

bensofficial:
Pushed 76473492 — `ApplicationsProvider` now derives the persisted `?topics=` value from raw `filters.topics` (`urlTopicsKey`) while keeping `adjustedFilters.topics` for the API request and refetch dep. The `showOnlyAssignedTopics` expansion no longer leaks into the URL.

This also unblocks the `application-comment-autosave` E2E failure: `ReviewApplicationPage` runs with `showOnlyAssignedTopics=true`, so the previous code was writing `?topics=NO_TOPIC,...` to the URL on `/applications`. Combined with the `location.search` preservation in the sidebar `onSelect` navigation, the URL became `/applications/<id>?topics=...`, which broke the test's `waitForURL('**/applications/${id}')` glob. With the URL clean by default, the glob matches again.

client/src/core/application/providers/ApplicationsProvider/ApplicationsProvider.tsx
Claudia-Anthropica: @bensofficial Looks good now, thanks. The latest change persists the raw `filters.topics` value and keeps the assigned-topic expansion out of the URL.

client/src/core/application/providers/ApplicationsProvider/ApplicationsProvider.tsx
Claudia-Anthropica: @bensofficial This is still open after `7647349`. The latest commit fixed the raw `topics` URL sync, but the reset effect still depends on `adjustedFilters`, so the page restored from `/applications?page=N` will still be reset when `useAllTopics()` resolves and expands the assigned-topic filter.

Claudia-Anthropica (changes requested):
@bensofficial The raw topics URL persistence issue is fixed, but the applications page reset is still open. The reset effect still depends on adjustedFilters, so loading assigned topics can wipe the page restored from the URL; please tie that reset to user-driven filter/sort changes instead.

bensofficial:
Pushed 3a9bb4b0 — two follow-ups based on your feedback:

1. **Page reset on async topic load**: dropped the standalone `useEffect(() => setPage(0), [sort, adjustedFilters])` entirely. The `didMountRef` guard was insufficient because `adjustedFilters` mutates after `useAllTopics()` resolves, firing the reset and wiping the URL-hydrated page. The `setFilters` / `setSort` context wrappers already call `setPage(0)` on user-driven changes, so async topic loading no longer touches `page`.

2. **URL sync writing default states**: `states=NOT_ASSESSED` was being written into the URL on `/applications` because `defaultStates` is non-empty. Added `defaultStatesKey` / `defaultTopicsKey` comparisons so the URL sync only writes when the current filter differs from the page-level defaults. The freshly loaded `/applications` stays at a clean URL, the `application-comment-autosave` test's `waitForURL('**/applications/${id}')` glob matches again, and user-changed filters still flow into the URL.

client/src/core/application/providers/ApplicationsProvider/ApplicationsProvider.tsx
Claudia-Anthropica: @bensofficial Looks good now, thanks. Dropping the async-driven reset and resetting only through the user-facing filter/sort setters preserves the restored page.

Claudia-Anthropica (approved):
@bensofficial All previously requested changes are addressed now. The deploy job is still pending, but the build, e2e, CodeQL, analysis, CodeRabbit, and Helios checks are passing; approving the follow-up.

PR #1160, opened 2026-07-02
Filter topics by supervisor name via shareable URL

bensofficial:
## Summary
- Adds a supervisor-scoped view of the public topic listing so supervisors can send a link that only shows their open topics.
- New `?supervisor=<full name>` query param and `/supervisor/:name` path route on the landing page. Matching is case-insensitive, whitespace-tolerant, and restricted to the SUPERVISOR role on a topic.
- Header switches to "Topics supervised by X" when the view is scoped.

## Details
- Backend: threads `supervisorName` through `TopicController` → `TopicService` → `TopicRepository.searchTopics` via an `EXISTS` clause on `topic_roles` joined with `users`.
- Client: adds `supervisorName` to `ITopicsFilters`, reads both the query param and the new path param on the landing page, and adds a `/supervisor/:supervisorName` route.
- Example shareable link: `https://thesis.aet.cit.tum.de/supervisor/Mateo%20de%20Mayo` or `https://thesis.aet.cit.tum.de/?supervisor=Mateo%20de%20Mayo`.

## Test plan
- [x] Server: 5 new integration tests in `TopicControllerTest.TopicSupervisorFiltering` cover basic filtering, case-insensitivity + trimming, unknown name → empty, multi-supervisor topic match, and that examiner-only role matches are excluded (all pass locally).
- [x] Server: existing `TopicServiceTest.getAll_ReturnsPageOfTopics` updated for the new arg (passes).
- [x] Client: `pnpm exec tsc --noEmit` and `pnpm exec eslint` on edited files are clean.
- [x] E2E: new Playwright spec `client/e2e/topic/topic-supervisor-filter.spec.ts` covers query param, path route, case-insensitivity, unknown name, and unfiltered baseline. Run via `./execute-e2e-local.sh`.
- [x] Manual: visit `/supervisor/Supervisor%20User` on the seeded dev instance and confirm only the ASE topics with that supervisor appear, and the header reads "Topics supervised by Supervisor User".

Closes #1159

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Topics can now be filtered by supervisor in the app, including a new supervisor-specific page route.
  * The page header now updates to show supervisor-related messaging when a supervisor filter is present.

* **Bug Fixes**
  * Supervisor filtering is case-insensitive and ignores extra whitespace.
  * Topics now correctly reflect the selected supervisor, including cases with no matches or multiple supervisors.

* **Tests**
[...]

PR #1163, opened 2026-07-02
Filter published theses by supervisor name

bensofficial:
## Summary
- Follow-up to #1160. The URL supervisor filter (`?supervisor=…` / `/supervisor/:name`) only reached the **open topics** view, so switching to **Published Theses** on `/supervisor/<name>` reset to the full list.
- Threads `supervisorName` through `PublishedThesisController` → `ThesisService.getAll` → `ThesisRepository.searchTheses` and forwards the value from `LandingPage` to `PublishedTheses`.
- The JPQL filter uses the same case-insensitive, whitespace-tolerant full-name match as the topic query and is restricted to the `SUPERVISOR` role (examiner-only role matches are excluded).

## Test plan
- [x] New `PublishedThesisControllerTest.GetPublishedThesesSupervisorFilter` covers (a) supervisor filtering isolates the matching thesis, and (b) examiner-only role matches are excluded.
- [x] Existing `ThesisServiceTest` mock updated for the new arg.
- [x] CI runs the full server test suite and client type check.
- [x] Manual: visit `/supervisor/<name>` on the deployed dev, toggle the view to **Published Theses**, confirm only that supervisor's finished theses appear.

## Summary by CodeRabbit

* **New Features**
  * Published theses can now be filtered by supervisor name on the landing page and published theses view.
  * Search results respect the selected supervisor, including values passed from the page route or search parameters.

* **Bug Fixes**
  * Improved matching for supervisor filters so names are handled consistently, regardless of case or extra spaces.
  * Fixed published thesis listings to exclude theses where the person is not actually listed as supervisor.


bensofficial:
@Claudia-Anthropica Can you please review the PR?

server/src/main/java/de/tum/cit/aet/thesis/thesis/repository/ThesisRepository.java
```diff
@@ -40,6 +40,12 @@ AND EXISTS (
 			)
 			AND (:states IS NULL OR t.state IN :states)
 			AND (:types IS NULL OR t.type IN :types)
+			AND (:supervisorName IS NULL OR EXISTS (
+				SELECT 1 FROM ThesisRole sr
+				WHERE sr.thesis.id = t.id
+					AND sr.id.role = de.tum.cit.aet.thesis.thesis.constants.ThesisRoleName.SUPERVISOR
+					AND LOWER(CONCAT(COALESCE(sr.user.firstName, ''), ' ', COALESCE(sr.user.lastName, ''))) = :supervisorName
```
Claudia-Anthropica: @bensofficial [medium] This does not quite match the topic supervisor filter yet: if a supervisor only has one stored name part, `COALESCE(firstName, empty) + space + COALESCE(lastName, empty)` leaves a leading or trailing space, while `supervisorName` was already trimmed in the service. Open topics would match that supervisor via `TRIM(...)`, but published theses would come back empty. Please trim the DB-side concatenated name here too.

In `server/src/main/java/de/tum/cit/aet/thesis/thesis/repository/ThesisRepository.java`, the supervisor-name filter compares an untrimmed concatenated first and last name against a trimmed parameter, so one-part supervisor names do not match published theses. Fix this by trimming the DB-side concatenated name before lowercasing or comparing, matching `TopicRepository` semantics, and add or adjust a test for a supervisor with a missing first or last name.

</details>

Claudia-Anthropica (changes requested):
@bensofficial The wiring from the landing page into the published theses endpoint looks good, but the new query doesn't quite match the existing topic supervisor filter for nullable name parts. See the inline comment.

server/src/main/java/de/tum/cit/aet/thesis/thesis/repository/ThesisRepository.java
Claudia-Anthropica: @bensofficial Looks good now, thanks.

Claudia-Anthropica (approved):
@bensofficial All feedback addressed, and the focused PublishedThesisControllerTest passes locally. CI still has pending jobs, but I don't see any code-review blockers.

PR #1167, opened 2026-07-07
Remove SBOM caching in repo

bensofficial:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved build reliability by generating required build artifacts earlier in the pipeline.
  * Reduced CI failures by removing checks that enforced stale artifact detection.

* **Chores**
  * Streamlined build and test workflows to simplify artifact handling.
  * Updated repository settings to keep generated build artifacts out of version control.


.github/workflows/build_docker.yml
```diff
@@ -66,6 +66,46 @@ jobs:
         if: ${{ steps.changed-files.outputs.any_changed == 'true' }}
         uses: docker/setup-buildx-action@v4
 
+      # Generate SBOMs on the runner before the Docker build so they land
```
Claudia-Anthropica, line 69: @bensofficial [high] This generation only happens in the Docker build job, but e2e starts the server with `./gradlew bootRun` before any client SBOM is generated. Since `client/sbom/bom.json` is no longer committed, the admin dependency page only gets the server SBOM; the current e2e failure in `source filter switches the visible counts` is exactly that. Generate the client SBOM before any server build/bootRun path that exercises the dependency page, or adjust that e2e setup so it does not expect both sources when the client SBOM is absent.

In `.github/workflows/e2e_tests.yml`, the server starts before `client/sbom/bom.json` exists after the checked-in SBOM was removed. Fix this by generating the client SBOM before `./gradlew bootRun` runs, using the direct generator path that avoids package lifecycle scripts, so the admin dependency e2e sees both server and client SBOMs.

</details>

.github/workflows/build_docker.yml
```diff
@@ -66,6 +66,46 @@ jobs:
...
+          cache: 'pnpm'
+          cache-dependency-path: 'client/pnpm-lock.yaml'
+
+      # `pnpm sbom` reads the lockfile directly, so we deliberately do NOT
+      # run `pnpm install` — skipping it both saves time and avoids
+      # running dependency lifecycle scripts. Corepack provides pnpm.
+      - name: Generate client SBOM
+        if: ${{ steps.changed-files.outputs.any_changed == 'true' }}
+        working-directory: ./client
+        run: pnpm run sbom
```
Claudia-Anthropica: @bensofficial [medium] `pnpm run sbom` still performs a dependency install on a clean runner before the script starts. The passing build log shows `Packages: +901`, `preinstall`, `postinstall`, and `prepare`, so this does not avoid lifecycle scripts; call `node scripts/generate-sbom.js` directly after Corepack is enabled instead.

In `.github/workflows/build_docker.yml`, the Generate client SBOM step runs `pnpm run sbom`, which triggers pnpm dependency installation and lifecycle scripts on clean runners. Fix this by invoking `node scripts/generate-sbom.js` directly with Corepack enabled so the nested `pnpm sbom --lockfile-only` still works without running `pnpm install` or package lifecycle scripts.

</details>

Claudia-Anthropica (changes requested):
@bensofficial The SBOM removal mostly works for the Docker build, but the e2e server path no longer gets a client SBOM and the PR is currently red because of that. I also left a note on the `pnpm run sbom` step because it still runs install/lifecycle scripts on a clean runner.

.github/workflows/build_docker.yml
Claudia-Anthropica: @bensofficial This is still the same issue: the current Docker workflow still runs `pnpm run sbom`, and the new e2e SBOM step does the same. Please call `node scripts/generate-sbom.js` directly in both places so a clean runner does not do an implicit install/lifecycle run before generating the SBOM.

Claudia-Anthropica (changes requested):
@bensofficial One of the previous blockers is fixed, but the `pnpm run sbom` lifecycle/install issue is still present in both SBOM-generation workflow steps. Keeping this as changes requested until those calls switch to the direct generator.

.github/workflows/build_docker.yml
Claudia-Anthropica: @bensofficial Looks good now, thanks.

Claudia-Anthropica (approved):
@bensofficial The previous SBOM blockers are addressed: e2e generates the client SBOM before server startup and both workflow paths call the generator directly. Approving the follow-up; CI was still pending when I checked.

PR #1199, opened 2026-07-26
Fix topic page applications list showing unrelated applications

bensofficial:
## Summary
- On the topic detail page, the applications section was rendering applications for other topics — specifically all "suggested" (no-topic, student-proposed) applications leaked in alongside the current topic's applications.
- Root cause: `ApplicationsProvider` hardcoded `includeSuggestedTopics: true`. Combined with the server clause `(:topics IS NULL OR a.topic.id IN :topics OR (:includeSuggestedTopics = true AND a.topic IS NULL))`, the topic filter was OR'd with `topic IS NULL`.
- Added a `defaultIncludeSuggestedTopics` prop (default `true` for backward compat) and pass `false` from `TopicPage` so the topic filter is strict.
- Removed the redundant "Topic" column from the table on the topic page — every row is now the current topic, so the column would just repeat the same title.

## Test plan
- [x] Open a topic detail page as a manager/examiner/supervisor of the topic's research group; confirm the applications table shows only applications for that topic.
- [x] Confirm no student-proposed (no-topic) applications appear in that list.
- [x] Confirm the "Topic" column is gone; State / Thesis Type / Student / Created At remain.
- [ ] Confirm other pages using `ApplicationsProvider` (main applications list, sidebar, etc.) still include suggested topics by default.

## Screenshots

Topic applications table without the redundant Topic column

## Summary by CodeRabbit

* **New Features**
  * Added configurable defaults for whether suggested topics are included in application filters.
  * Topic management now excludes suggested topics by default and displays only relevant application details.

* **Bug Fixes**
  * Improved application filtering and table presentation in topic management.


Claudia-Anthropica (approved):
@bensofficial The strict topic filter works as intended, existing provider callers keep their previous behavior, and the simplified table looks good. Approving as-is.

bensofficial:
Tested on Dev. Works.

PR #1200, opened 2026-07-26
Remove NO_TOPIC sentinel from application topic filter

bensofficial:
## Summary
- On the Review Applications overview, the topic filter auto-populated a raw `NO_TOPIC` pill for reviewers because a sentinel was added to `filters.topics`. On the server the sentinel had no functional effect — null-topic applications are only surfaced via the separate `includeSuggestedTopics` query param (see `ApplicationRepository.searchApplications`, `a.topic.id IN :topics OR (:includeSuggestedTopics = true AND a.topic IS NULL)`), so the two overlapped in intent while only one of them actually did anything.
- Removed the `NO_TOPIC` sentinel from `ApplicationsProvider` (both the auto-populate branch and the zero-UUID request-time conversion). `showOnlyAssignedTopics` still pre-selects the reviewer's assigned topics; whether null-topic applications appear is now controlled solely by the "Include suggested topics" switch.
- Moved the "Include suggested topics" switch into the Topic column, directly under the Topic MultiSelect, so the relationship between the two controls is visible at a glance.

## Behaviour note
A reviewer who is assigned to **no** topics used to see only null-topic (student-proposed) applications by default because the sentinel filtered everything else out. After this change, an unassigned reviewer sees all applications in their research group by default, and can still narrow with the switch. This seemed like the more reasonable default, and the change stays inside the "Review Applications" surface.

## Test plan
- [x] `pnpm exec eslint` on the two changed files
- [x] `pnpm exec tsc --noEmit`
- [x] `pnpm exec vitest run src/core/application/components/ApplicationsFilters` (3/3 pass)
- [x] Manual: on `/applications`, verify no `NO_TOPIC` pill in the Topic filter and that toggling "Include suggested topics" hides/shows student-proposed applications

## Screenshots

Review Applications topic filter


Claudia-Anthropica (approved):
@bensofficial The sentinel removal matches the server query semantics, and the colocated switch works cleanly in the running application. Nice work—approving as-is.

bensofficial:
Tested locally. Works.

PR #1205, opened 2026-07-26
Accept Resource in AbstractAutoFillService, matching PDF extractor

bensofficial:
## Summary
- Mirror the `MultipartFile` / `Resource` dual-entry pattern of `feedback/PdfService` in `AbstractAutoFillService`. The existing `MultipartFile` overload keeps working; a new `process(Thesis, Resource)` overload lets callers that only hold a stored-file `Resource` (e.g. the AI review pipeline's `loadPdfResource`) run abstract auto-fill against the persisted PDF without needing the original multipart.
- Both overloads route through a private byte-based path so the extractor call site stays a single line, matching `PdfService`'s shape.
- Adds two tests covering the `Resource` overload: a confident-fill case from a stored PDF and an unreadable-resource no-op.

Base branch is `feat/persist-ai-feedback` (not `develop`) because the AI review pipeline is the motivating consumer for the `Resource` overload.

## Test plan
- [ ] `cd server && ./gradlew test --tests AbstractAutoFillServiceTest` (all 11 tests pass locally)
- [ ] Verify existing `ThesisService.uploadProposal` / `uploadThesisFile` callers still pass a `MultipartFile` and continue to work unchanged
- [ ] Confirm no callers were broken by the overload resolution

### DIFF
--- a/server/src/main/java/de/tum/cit/aet/thesis/thesis/service/AbstractAutoFillService.java
+++ b/server/src/main/java/de/tum/cit/aet/thesis/thesis/service/AbstractAutoFillService.java
@@ -3,11 +3,13 @@ package de.tum.cit.aet.thesis.thesis.service;
 import de.tum.cit.aet.thesis.core.utility.AbstractExtractor;
 import de.tum.cit.aet.thesis.thesis.constants.ThesisAbstractSource;
 import de.tum.cit.aet.thesis.thesis.entity.Thesis;
-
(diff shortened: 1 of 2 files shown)


bensofficial:
Wrong direction — the ask is for AI-driven abstract extraction (LLM-enhanced OCR/reading), not a Resource overload of the existing static extractor. Closing this and starting over on a new branch.

PR #1210, opened 2026-07-26
Add AI-driven abstract extractor using PdfService

bensofficial:
## Summary
- Introduce `AiAbstractExtractor`, a Spring service (conditional on AI features) that identifies the thesis / proposal abstract through an LLM instead of the deterministic iText / font-size heuristics.
- Text comes from the existing `feedback/PdfService.extractTextFromPdf`; front-matter pages (max 10) are fenced with a strict security prompt and handed to `ChatClient`, which returns CONFIDENT / UNCERTAIN / NONE plus paragraph HTML — matching the existing `AbstractExtractor.Result` shape so `AbstractAutoFillService.apply` is unchanged.
- `AbstractAutoFillService` now injects `ObjectProvider<AiAbstractExtractor>` and uses the LLM path when the bean is present. On deployments without AI features the bean is absent and the service transparently falls back to the static `AbstractExtractor` so uploads still auto-fill.
- The LLM prompt asks the model to reconstruct paragraphs, rejoin hyphenated line-breaks, and correct obvious OCR-style artifacts — the "AI-enhanced OCR" behaviour that the deterministic extractor cannot do.

Base branch is `feat/persist-ai-feedback` (not `develop`) because the new service reuses `feedback/PdfService` which lives on that branch.

## Test plan
- [x] `cd server && ./gradlew test --tests AbstractAutoFillServiceTest --tests AiAbstractExtractorTest` — all 16 tests pass locally
- [ ] Manual: upload a scanned/image-heavy proposal PDF and verify the abstract is auto-filled cleanly (needs a running LLM)
- [ ] Manual: run with `thesis-management.ai.enabled=false` and confirm the deterministic fallback still works


server/src/main/java/de/tum/cit/aet/thesis/thesis/service/AbstractAutoFillService.java
```diff
@@ -29,12 +41,22 @@ public class AbstractAutoFillService {
...
+			apply(thesis, extract(file));
 		} catch (Exception e) {
 			log.warn("Abstract extraction failed for thesis {}: {}", thesis.getId(), e.getMessage());
 		}
 	}
 
+	private AbstractExtractor.Result extract(MultipartFile file) throws java.io.IOException {
+		AiAbstractExtractor aiExtractor = aiAbstractExtractorProvider.getIfAvailable();
+		if (aiExtractor != null) {
+			return aiExtractor.extract(file);
```
Claudia-Anthropica, line 53: @bensofficial [high] This is now a blocking network call, but both callers (`uploadProposal` and `uploadThesisFile`) are `@Transactional` and invoke it after repository work. A slow model can therefore hold a database transaction/connection for the full LLM latency and exhaust the pool under concurrent uploads; move extraction outside the transaction or run it after commit with a separate persistence step.

In `server/src/main/java/de/tum/cit/aet/thesis/thesis/service/AbstractAutoFillService.java`, AI extraction is called synchronously from transactional upload methods in `ThesisService`. Refactor the upload flow so the remote LLM call runs outside the database transaction, then persist the extraction result in a short separate transaction while preserving best-effort behavior.

</details>

server/src/main/java/de/tum/cit/aet/thesis/thesis/service/AiAbstractExtractor.java
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	/**
+	 * Extracts the abstract from the uploaded PDF.
+	 *
+	 * @param file the uploaded proposal or thesis PDF
+	 * @return the extraction result — the caller's apply logic decides how it lands on the thesis
+	 */
+	public AbstractExtractor.Result extract(MultipartFile file) {
+		List<String> pages = pdfService.extractTextFromPdf(file);
```
Claudia-Anthropica, line 107: @bensofficial [high] `extractTextFromPdf` only reads embedded PDF text; it performs no OCR and this request sends no page images. An image-only scan therefore produces blank page strings, so the model sees only page markers and cannot satisfy the PR's scanned/image-heavy use case. Feed the first ten rendered pages to a vision-capable model or run OCR first, and cover an image-only PDF in a test.

In `server/src/main/java/de/tum/cit/aet/thesis/thesis/service/AiAbstractExtractor.java`, image-only scanned PDFs provide no usable input because only embedded text is sent to the model. Add a real OCR or first-ten-page vision input path and an integration-style test using a PDF whose abstract exists only as an image.

</details>

server/src/main/java/de/tum/cit/aet/thesis/thesis/service/AiAbstractExtractor.java
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	private static AbstractExtractor.Result sanitize(AbstractExtractor.Result result) {
+		if (result == null || result.confidence() == null) {
+			// Defensive: an LLM that fails structured decoding gives us nothing to apply.
+			log.debug("AI abstract extractor returned no structured result");
+			return new AbstractExtractor.Result(AbstractExtractor.Confidence.NONE, "");
+		}
+		String html = result.html() != null ? result.html() : "";
+		return new AbstractExtractor.Result(result.confidence(), html);
```
Claudia-Anthropica: @bensofficial [medium] The model's `confidence` and HTML are accepted verbatim, so a `CONFIDENT` response shorter than 50 characters, longer than 2000, or containing non-paragraph markup is silently stored even though the prompt forbids it. Prompt rules are not validation; enforce the length/markup contract here and downgrade invalid confident results to `UNCERTAIN` (or `NONE` when empty).

In `server/src/main/java/de/tum/cit/aet/thesis/thesis/service/AiAbstractExtractor.java`, `sanitize` trusts model confidence and HTML without enforcing the prompt's plausibility rules. Validate plain-text length and paragraph-only safe HTML in code, downgrade invalid `CONFIDENT` results appropriately, and add tests for short, overlong, and disallowed-markup responses.

</details>

Claudia-Anthropica (changes requested):
@bensofficial The overall extraction flow is well structured, and the focused tests pass. Before merging, the LLM call needs to leave the upload transaction, scanned pages need an actual visual/OCR path, and the model response contract needs server-side enforcement; see the inline comments.

server/src/main/java/de/tum/cit/aet/thesis/thesis/service/AbstractAutoFillService.java
bensofficial: Fixed in 56038c7e. The LLM call is now out of the upload transaction entirely.

`uploadProposal` / `uploadThesisFile` keep their `@Transactional` DB work but no longer call the extractor. Instead the controller invokes a new **non-transactional** `ThesisService.autoFillAbstractFromUpload(thesis, file)` *after* the upload transaction has committed. That method runs the extraction with no transaction active and then persists the result via a single `thesisRepository.save(...)` (a short per-call transaction). So the model latency can never hold a DB connection open, and it stays best-effort — any failure is swallowed and logged, never surfacing to the client.

server/src/main/java/de/tum/cit/aet/thesis/thesis/service/AiAbstractExtractor.java
bensofficial: Fixed in 56038c7e. There is now a real vision path for scanned / image-only PDFs.

`extract` first pulls embedded text; if the combined front-matter text is below `MIN_EMBEDDED_TEXT_CHARS` (i.e. an image-only scan produces essentially no text), it renders the first `MAX_PAGES` pages via the new `PdfService.extractImagesFromPdf(file, maxPages)` and attaches them to the user message so a vision-capable model reads the abstract from the images. Born-digital PDFs with real embedded text keep the cheap text-only path (no rendering cost, which also helps the latency concern above). The system prompt now tells the model to read from the page images when the text is empty or garbled.

On testing: a true end-to-end assertion needs a live vision model, which we can't run in CI. The unit tests instead pin the branch behaviour — `extract_imageOnlyScan_rendersPagesAndSendsImagesToVisionModel` feeds blank page text, asserts the pages are rendered and that the rendered images are attached to the chat request, and `extract_textRichPdf_usesTextOnlyPathWithoutRenderingImages` asserts images are *not* rendered when embedded text is present.

server/src/main/java/de/tum/cit/aet/thesis/thesis/service/AiAbstractExtractor.java
bensofficial: Fixed in 56038c7e. `sanitize` now enforces the contract in code rather than trusting the prompt.

It computes the plain-text length (tags stripped, whitespace collapsed) and checks the markup:
- empty result → `NONE`;
- a `CONFIDENT` result whose plain text is shorter than `MIN_CONFIDENT_LENGTH` (50), longer than `MAX_CONFIDENT_LENGTH` (2000), or that contains any tag other than `<p>` / `</p>` → downgraded to `UNCERTAIN`, so it is offered as a suggestion for the student to confirm instead of being auto-filled silently.

`MIN/MAX_CONFIDENT_LENGTH` mirror the deterministic `AbstractExtractor` bounds so both paths agree on plausibility. New tests cover the short, overlong, disallowed-markup, and empty-html cases, plus a plausible result that stays `CONFIDENT`.

server/src/main/java/de/tum/cit/aet/thesis/thesis/service/ThesisService.java
```diff
@@ -724,12 +725,33 @@ public Thesis uploadThesisFile(Thesis thesis, String type, MultipartFile file) {
...
+	 * result. Best-effort — any failure is swallowed so it can never break the completed upload.
+	 *
+	 * @param thesis the thesis whose abstract should be auto-filled (mutated in place)
+	 * @param file the uploaded proposal or thesis PDF
+	 * @return the thesis, saved when extraction produced a change and returned as-is otherwise
+	 */
+	public Thesis autoFillAbstractFromUpload(Thesis thesis, MultipartFile file) {
+		try {
 			abstractAutoFillService.process(thesis, file);
+			return thesisRepository.save(thesis);
```
Claudia-Anthropica, line 748: @bensofficial [high] `thesis` is the detached snapshot returned before the potentially long extraction. Because `Thesis` has no `@Version`, `save(thesis)` merges its stale scalar fields after the LLM returns, so a concurrent info, abstract, or state update can be silently overwritten; this save also runs when extraction fails because `process` swallows the exception. Extract first, then reload the current entity in a short transaction and apply the result to that managed instance, with optimistic or conditional protection.

In `server/src/main/java/de/tum/cit/aet/thesis/thesis/service/ThesisService.java`, `autoFillAbstractFromUpload` merges a stale detached `Thesis` after LLM latency, allowing concurrent changes to be overwritten. Refactor extraction to return a result, then reload and apply it in a short separate transaction with optimistic or conditional protection before persisting.

</details>

Claudia-Anthropica (changes requested):
@bensofficial The response contract is now enforced and the LLM no longer runs inside the upload transaction. However, scanned PDFs still fail with the repository's AI-enabled text-only development model because the new branch unconditionally attaches media, and the post-LLM save can silently overwrite concurrent thesis changes from a stale detached entity. These blocking issues need correction before approval; the snapshot has no required checks, with Helios successful and CodeQL queued as non-required.

server/src/main/java/de/tum/cit/aet/thesis/thesis/service/AbstractAutoFillService.java
Claudia-Anthropica: @bensofficial Thanks, this concern is fixed: extraction now runs after the upload transaction has committed.

server/src/main/java/de/tum/cit/aet/thesis/thesis/service/AiAbstractExtractor.java
Claudia-Anthropica: @bensofficial The image branch is present, but this remains incomplete for AI-enabled text-only configurations: `application-dev.yml` enables `openai/gpt-oss-120b`, while `AiAbstractExtractor` always attaches rendered images for sparse PDFs. Detect vision support and provide an OCR/text fallback when images cannot be sent, with coverage for both capability branches.

PR #1190, opened 2026-07-14
Redesign thesis overview page with sticky process nav

bensofficial:
## Summary

- **Configuration** is now a dedicated page at `/theses/:id/configuration`, gated to supervisors and examiners of the thesis (route-level check against `admin/advisor/supervisor` groups + in-page check against `access.supervisor`). A "Configuration" button in the thesis header (visible only to supervisors) links to it.
- **Info + Involved Persons + Comments** are merged into a single "Overview" accordion with three items to reduce top-level clutter.
- **Sticky process navbar** on the thesis page with six steps (Overview → Proposal → Thesis → Presentation → Assessment → Grade) and `IntersectionObserver`-based scroll-spy. Clicking a step smooth-scrolls to the section; steps for hidden/non-applicable sections are omitted.
- **Non-applicable sections hidden**: Proposal hides when the thesis has moved past the proposal phase without a file; Presentation hides when there are none and the viewer can't create one.

The three old section files (`ThesisInfoSection`, `ThesisStudentInfoSection`, `ThesisSupervisorCommentsSection`) are deleted; the two info sub-components (`AbstractSuggestionModal`, `DownloadAllFilesButton`) are moved into the new `ThesisOverviewSection/components/` directory.

## Test plan

- [ ] Log in as a student on a thesis in `WRITING` state → verify sticky navbar appears with Overview/Thesis (and Proposal only if a proposal file exists); Configuration button is **not** shown in header.
- [ ] Log in as a supervisor → "Configuration" button appears in header; clicking it lands on `/theses/:id/configuration` with the existing edit form and a "Back to Thesis" link.
- [ ] As a non-supervisor, visit `/theses/:id/configuration` directly → redirects to `/theses/:id`.
- [ ] Scroll the thesis page → active navbar step updates to match the section in view.
- [ ] Click a nav step → smooth-scrolls to that section with the section top aligned below the navbar.
- [ ] Thesis in `ASSESSED` state → "Grade" step appears in the navbar; "Proposal" step is hidden if no proposal was ever uploaded.
- [ ] Thesis in `SUBMITTED` state viewed by a student → "Assessment" step remains hidden (existing behavior preserved).
- [ ] `pnpm exec tsc --noEmit` and `pnpm exec eslint src/` both pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
[...]

PR #1242, opened 2026-08-06
Fix process nav highlight snapping back to Overview on click

bensofficial:
## Problem

On the redesigned thesis overview page (#1190), clicking a step in the sticky process nav scrolls to that section but the highlighted ("active") pill often snaps back to **Overview**.

Cause: the active highlight is driven entirely by the scroll spy, which highlights the topmost visible section. When the sections above the target are collapsed/short, `scrollIntoView` can't bring the target all the way to the top (not enough content below it), so Overview stays at the top of the viewport and keeps the highlight — even though the user clicked, e.g., **Thesis**. A click had no way to assert the active section itself.

## Fix

`useActiveSection` now returns `[activeId, setActiveSection]`. On click, the nav:

1. Calls `setActiveSection(id)` to highlight the clicked section immediately, and
2. **Locks** the scroll spy so its `IntersectionObserver`/scroll re-evaluation can't override the choice.

The lock releases on the next user-initiated scroll (`wheel`, `touchmove`, or a scroll key — arrows/PageUp-Down/Home/End/Space), after which scroll-spy tracking resumes exactly as before. Plain typing in inputs is ignored so the lock doesn't drop unexpectedly.

Net effect: clicking a nav step highlights that step and keeps it highlighted until the user scrolls away, matching intent regardless of how far the page can actually scroll.

## Testing

- `pnpm exec tsc --noEmit` — clean
- `pnpm exec eslint` on both changed files — clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved thesis section navigation so the selected step is highlighted immediately when clicked.
  * Prevented scroll-based navigation updates from overriding the user’s selection until users resume scrolling or use touch, pointer, wheel, or keyboard navigation.
  * Improved active-section tracking within scrollable areas, including at their boundaries.

### DIFF
--- a/client/src/thesis/pages/ThesisPage/hooks/useActiveSection.ts
+++ b/client/src/thesis/pages/ThesisPage/hooks/useActiveSection.ts
@@ -1,13 +1,44 @@
-import { useEffect, useState } from 'react'
+import { useCallback, useEffect, useRef, useState } from 'react'
+
+// Keys that scroll the page — a keydown from one of these counts as the user
+// taking over navigation and releases a click lock (see below). Plain typing in
+// an input must not release it, so unrelated keys are ignored.
+const SCROLL_KEYS = new Set([
+  'ArrowUp',
+  'ArrowDown',
+  'PageUp',
+  'PageDown',
+  'Home',
+  'End',
+  ' ',
+  'Spacebar',
+])
 
 /**
  * Tracks which of the given section anchor ids is currently in view. Uses
  * IntersectionObserver with a top offset so a section counts as "active" once
  * it scrolls above the sticky navbar rather than only when centered.
+ *
+ * Returns the active id plus a `setActiveSection` setter. Call the setter when
+ * the user clicks a nav item: it highlights that section immediately and locks
+ * the scroll spy so it can't snap the highlight back. Without the lock, clicking
+ * a nav item on a short/collapsed page (where the target can't scroll all the
+ * way to the top) would leave whatever sits at the top — usually "Overview" —
+ * highlighted, which is misleading. The lock is released as soon as the user
+ * scrolls the page themselves, after which the scroll spy resumes tracking.
  */
-export function useActiveSection(sectionIds: string[], topOffset = 100): string | null {
+export function useActiveSection(
+  sectionIds: string[],
+  topOffset = 100,
+): [string | null, (id: string) => void] {
   const [activeId, setActiveId] = useState<string | null>(sectionIds[0] ?? null)
   const key = sectionIds.join('|')
+  const lockedRef = useRef(false)
+
+  const setActiveSection = useCallback((id: string) => {
+    lockedRef.current = true
+    setActiveId(id)
+  }, [])
 
   useEffect(() => {
     if (sectionIds.length === 0) {
@@ -44,6 +75,12 @@ export function useActiveSection(sectionIds: string[], topOffset = 100): string 
     }
 
     const pickActive = () => {
+      // A click lock pins the highlight to the user's chosen section until they
+      // scroll the page themselves, so don't override it here.
+      if (lockedRef.current) {
+        return
+      }
+
       if (isScrolledToBottom()) {
         setActiveId(elements[elements.length - 1].id)
         return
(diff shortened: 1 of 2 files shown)


client/src/thesis/pages/ThesisPage/hooks/useActiveSection.ts
```diff
@@ -84,12 +121,32 @@ export function useActiveSection(sectionIds: string[], topOffset = 100): string
...
+        pickActive()
+      }
+    }
+    const onKeyDown = (event: KeyboardEvent) => {
+      if (SCROLL_KEYS.has(event.key)) {
+        releaseLock()
+      }
+    }
+    window.addEventListener('wheel', releaseLock, { passive: true, capture: true })
+    window.addEventListener('touchmove', releaseLock, { passive: true, capture: true })
```
Claudia-Anthropica: @bensofficial [medium] This global `touchmove` handler also fires when a user swipes the nav's own horizontal `ScrollArea`. That gesture does not move the page, but it drops the lock and immediately calls `pickActive()`, so the clicked pill can snap back merely while the user is revealing another nav item on mobile.

In `client/src/thesis/pages/ThesisPage/hooks/useActiveSection.ts`, the global touchmove listener releases the click lock for horizontal process-nav swipes even when the page has not scrolled. Gate lock release on an observed change to the relevant vertical scroll position so horizontal nav gestures leave the lock intact.

</details>

Claudia-Anthropica (changes requested):
@bensofficial The click override addresses the original smooth-scroll race, but its global release handlers also clear the lock for gestures that do not scroll the page. In particular, horizontally swiping the process nav can reproduce the highlight snap-back on mobile, so the release condition needs tightening.

client/src/thesis/pages/ThesisPage/hooks/useActiveSection.ts
```diff
@@ -84,12 +121,32 @@ export function useActiveSection(sectionIds: string[], topOffset = 100): string
...
+    const releaseLock = () => {
+      if (lockedRef.current) {
+        lockedRef.current = false
+        pickActive()
+      }
+    }
+    const onKeyDown = (event: KeyboardEvent) => {
+      if (SCROLL_KEYS.has(event.key)) {
+        releaseLock()
+      }
```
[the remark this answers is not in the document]
bensofficial: Valid — fixed in f5ded481. Added an `isEditableTarget` guard so a keydown inside an `<input>`/`<textarea>`/`<select>`/contenteditable never arms a lock release. On top of that, release now also requires an observed vertical-scroll change, so a caret/Space key that doesn't scroll the page can't release the lock regardless.

client/src/thesis/pages/ThesisPage/hooks/useActiveSection.ts
bensofficial: Good catch — fixed in f5ded481. A gesture (wheel/touchmove/scroll-key) now only *arms* the release; the lock is actually released in the `scroll` handler, and only when the combined vertical scroll offset (`window.scrollY + scrollAncestor.scrollTop`) actually changes. A horizontal swipe of the nav's `ScrollArea` produces no vertical change, so the lock — and the clicked pill — stay put. This also keeps the programmatic smooth scroll (no gesture) from self-releasing.

client/src/thesis/pages/ThesisPage/hooks/useActiveSection.ts
```diff
@@ -79,17 +142,48 @@ export function useActiveSection(sectionIds: string[], topOffset = 100): string
...
+        armedRef.current = false
+      }
+      lastScroll = current
+      // Fallback for the "scrolled past the observer's active band" case for
+      // short final sections that never trip the intersection.
+      pickActive()
+    }
     window.addEventListener('scroll', onScroll, { passive: true, capture: true })
 
+    const arm = () => {
```
Claudia-Anthropica, line 169: @bensofficial [medium] The release can only be armed by wheel, touchmove, or keyboard input. Dragging either the window's or the detected ancestor's vertical scrollbar changes `current` without setting `armedRef`, so `lockedRef` remains true and the active pill no longer follows the user's manual scroll.

In `client/src/thesis/pages/ThesisPage/hooks/useActiveSection.ts`, vertical scrollbar drags change the page position without arming the click-lock release. Add a reliable pointer-driven scrollbar path for both the window and `scrollAncestor` that releases the lock after the vertical offset changes while leaving programmatic smooth scrolling locked.

</details>

Claudia-Anthropica (changes requested):
@bensofficial The vertical-offset gate addresses the reported horizontal-swipe case, and editable controls are now guarded. However, dragging a vertical scrollbar never arms the release gate, so the clicked pill remains pinned while the user manually scrolls to another section.

client/src/thesis/pages/ThesisPage/hooks/useActiveSection.ts
bensofficial: Fixed in 4553c3bd. `pointerdown` is now also an arming signal, so dragging the window's or the scroll ancestor's vertical scrollbar arms the release; the drag then changes the vertical offset and the lock is released, letting the pill follow the manual scroll again. Arming on `pointerdown` is safe because release still requires an actual vertical-offset change — a plain content click arms but never releases — and the click that kicks off a programmatic scroll resets `armedRef` before that scroll starts, so smooth scrolling stays locked.

Claudia-Anthropica (approved):
@bensofficial The scrollbar-drag path now arms the release while the vertical-offset gate continues to protect horizontal gestures and programmatic smooth scrolling. The prior editable-control and manual-scroll concerns are resolved, and the available required CI evidence has no failures.

Claudia-Anthropica (approved):
@bensofficial The exact head now checks the detected scroll ancestor exclusively for bottom detection, preventing the window fallback from falsely selecting the final section. The prior editable-input, horizontal-gesture, and scrollbar-scroll paths remain addressed, and the captured build, test, and E2E checks succeeded; the non-required deployment job was still waiting.

Claudia-Anthropica (approved):
@bensofficial The new commit correctly limits accordion auto-expansion to fully collapsed roots, preserving the selected Files/Comments panel while still revealing collapsed section content. The previously reviewed click-lock safeguards remain intact, and the exact-head client/server build and CodeQL checks succeeded; the non-required test and E2E jobs were still in progress in the snapshot.

PR #1308, opened 2026-09-08
Thesis abstract extractor fixes

mohamedalaaser:
## Summary
- Preserve real paragraph breaks when extracting abstracts from uploaded PDFs (modest gaps, short last lines, and indents).
- Keep compound hyphens that PDFs encode as soft hyphens (`role-sensitive` instead of `rolesensitive`), while still rejoining line-end hyphenation.

## Test plan
- [ ] Upload a multi-paragraph thesis PDF with a blank abstract and confirm paragraphs stay separated
- [ ] Confirm compounds like `role-sensitive` / `guideline-based` keep their hyphens
- [ ] Confirm line-broken words like `auto-` + `mated` still rejoin to `automated`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PDF paragraph detection for modest vertical gaps and sentence-ending short lines.
  * Prevented ordinary full-width line wraps from being incorrectly split into separate paragraphs.
  * Standardized soft hyphens as visible hard hyphens, including within compound words.

* **Tests**
  * Added coverage for paragraph splitting, line-wrap preservation, and hyphen normalization.

### DIFF
--- a/server/src/main/java/de/tum/cit/aet/thesis/core/utility/AbstractExtractor.java
+++ b/server/src/main/java/de/tum/cit/aet/thesis/core/utility/AbstractExtractor.java
@@ -35,10 +35,19 @@ public final class AbstractExtractor {
 	private static final int MAX_PAGES = 12;
 	/** Baselines within this many points are considered the same text line. */
 	private static final float LINE_Y_TOLERANCE = 3f;
-	/** A new paragraph starts when the vertical gap exceeds this multiple of the median line gap. */
-	private static final float PARAGRAPH_GAP_FACTOR = 1.5f;
+	/**
+	 * A new paragraph starts when the vertical gap exceeds this multiple of the typical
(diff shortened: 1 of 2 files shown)


Claudia-Anthropica (changes requested):
@mohamedalaaser [medium] I confirmed the unresolved CodeRabbit thread at `AbstractExtractor.java:398` remains valid: the short-line heuristic can split an ordinary ragged-right sentence wrap despite normal spacing and no indentation. The full-width regression does not cover the challenged case of a short sentence-ending wrap followed by a capitalized word. The captured CodeRabbit and Helios checks are successful.

Claudia-Anthropica (approved):
@mohamedalaaser The current head resolves the prior paragraph-splitting defect by requiring a modest vertical-gap cue before the short-line heuristic, and the new regression covers the normal-leading ragged-right case. The soft-hyphen handling and its extraction coverage are consistent with the intended PDF normalization. Helios is successful; the non-required CodeRabbit re-review remains pending.

